PR #10, opened 2025-07-10
added fmt, clippy, nextest to github ci

### LINKED ISSUE #4: [Task] Add minimal CI
### Background

Add support for CI:
- clippy
- fmt
- test

@netrome do we need anything else?

### User Story

This will certainly improve the development process

### Acceptance Criteria

A github workflow is added and tested manually

We can take most bits from [mpc ci](https://github.com/near/mpc/blob/main/.github/workflows/ci.yml)

gilcu3:
Fixes #4 

Notice that on the current main the checks fail for trivial reasons.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,44 @@
+name: CI
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+on:
+  workflow_dispatch:
+  pull_request:
+  merge_group:
+
+jobs:
+  ci-tests:
+    name: "Run tests"
+    runs-on: warp-ubuntu-2204-x64-8x
+    timeout-minutes: 60
+    permissions:
+      contents: read
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Cache Rust dependencies
+        uses: WarpBuilds/rust-cache@v2
+        with:
+          cache-on-failure: true
+          cache-all-crates: true
+
+      - name: Run clippy fmt
+        run: |
+          CARGO_TARGET_DIR="target/clippy" \
+          RUSTFLAGS="-D warnings" \
+          cargo clippy --all-features --all-targets --locked
+
+      - name: Run Cargo fmt
+        run: |
+          cargo fmt -- --check
+
+      - name: Install cargo-nextest
+        run: cargo install cargo-nextest
+
+      - name: Run cargo-nextest
+        run: cargo nextest run --release --locked --all-features


netrome:
> Notice that on the current main the checks fail for trivial reasons. @netrome should I fix them here?

Yes please do. As long as they are trivial/non-functional changes.

netrome:
> > Yes please do. As long as they are trivial/non-functional changes.
> 
> Taking a deeper look, I think many of them might be quite subtle and required proper review and tests (and there are quite a few), so I think it might be better to make that part of another PR. Do you agree?

Had a look. I see a lot of trivial ones. I assume many could be fixed with `cargo clippy --fix`. Some subtle ones such as the complex types lint would require some work and extra review.

I'd recommend:
1. Fix all simple/trivial ones.
2. Add #[allow(...)] for the more complex linter errors and add follow-up issues to address them.

gilcu3:
@SimonRastikian the file in `examples/network-benches.rs`  is making the CI fail. Afaik that file is totally outdated, so should we remove it and add proper examples later?

gilcu3:
This PR should be ready now

examples/network-benches.rs
netrome, line 1: Why is this example removed?

src/ecdsa/math.rs
```diff
@@ -104,6 +104,11 @@ impl<C: CSCurve> Polynomial<C> {
     pub fn len(&self) -> usize {
         self.coefficients.len()
     }
+
+    /// Required by https://rust-lang.github.io/rust-clippy/master/index.html#len_without_is_empty
+    pub fn is_empty(&self) -> bool {
+        self.coefficients.is_empty()
+    }
```
netrome, line 111: Makes sense.

src/ecdsa/triples/multiplication.rs
```diff
@@ -289,10 +289,11 @@ mod test {
                     .collect()
             });
 
+        #[allow(clippy::type_complexity)]
```
netrome, line 292: Would be nice with a follow-up issue to refactor these types. Definitely less trivial though as it requires some thinking and proper type design.

netrome (changes requested):
I don't understand why the benches example was removed? As a follow-up we should probably use criterion or divan for these benches (or hyperfine if it's more e2e - I don't know the details of our benchmark implementations). But in the meantime I'd expect them to stay relatively unmodified (just need to update the package name and probably fix some linter errors there as well).

examples/network-benches.rs
gilcu3: Afaik the example no longer corresponds with the rest of the code. Certainly worth it bringing it back, I will create a follow up issue on that once this is merged referencing the commit removing it.

PR #13, opened 2025-07-15
chore: update outdated dependencies and remove unused

### LINKED ISSUE #11: [Task] Clean dependencies
### Background

A few dependencies are a couple of major versions outdated, others seem unused.
- Removing `criterion` should solve this [security issue](https://github.com/near/threshold-signatures/security/dependabot/1)

### User Story

Avoid bundling unnecessary or outdated dependencies.

### Acceptance Criteria

- Dependencies are up-to-date and organized according to function in `Cargo.toml`
- Unnecessary dependencies are removed

_No response_

gilcu3:
Closes #11 

One dependency (`rand_core`) was not trivial to update due to quite a few API changes and the current version in use being quite old. I can give it a try if that's deemed required.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ byteorder = "1.5.0"
 digest = "0.10.7"
 ecdsa = { version = "0.16.8", features = ["digest", "hazmat"] }
 elliptic-curve = { version = "0.13.5", features = ["serde"] }
-event-listener = "2.5.3"
+# event-listener = "5.4.0"
 frost-core = { version = "2.1.0", default-features = false, features = ["serialization", "std", "internals"] }
 frost-ed25519 = { version = "2.1.0", default-features = false, features = ["serialization", "std"] }
 frost-secp256k1 = { version = "2.1.0", default-features = false, features = ["serialization", "std"] }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -21,20 +21,21 @@ itertools = "0.14.0"
...
 
 [dev-dependencies]
-criterion = "0.4"
-easy-parallel = "3.2.0"
-haisou-chan = { git = "https://github.com/cronokirby/haisou-chan", rev = "d28c46e51acfcb818236caae293f6e56dff41ad2" }
-structopt = "0.3.26"
+# criterion = "0.4"
+# easy-parallel = "3.2.0"
+# haisou-chan = { git = "https://github.com/cronokirby/haisou-chan", rev = "d28c46e51acfcb818236caae293f6e56dff41ad2" }
+# structopt = "0.3.26"
```
netrome, line 38: Why not completely remove these?

Cargo.toml
```diff
@@ -12,7 +12,7 @@ byteorder = "1.5.0"
 digest = "0.10.7"
 ecdsa = { version = "0.16.8", features = ["digest", "hazmat"] }
 elliptic-curve = { version = "0.13.5", features = ["serde"] }
-event-listener = "2.5.3"
+# event-listener = "5.4.0"
```
netrome, line 15: Why is this one commented out and not removed?

netrome (changes requested):
Thanks for doing this. I'd like not to retain these commented out. It's better to remove them, or if for any reason you'd like to keep them commented out - please add a comment explaining why.

gilcu3:
> Thanks for doing this. I'd like not to retain these commented out. It's better to remove them, or if for any reason you'd like to keep them commented out - please add a comment explaining why.

I was split between leaving them commented to be probably enabled later, or totally removing them. I am fine with the latter.

Cargo.toml
gilcu3: done

netrome:
> I was split between leaving them commented to be probably enabled later, or totally removing them. I am fine with the latter.

Yeah they still exist in the version history if for any reason we need to look up the exact version in the future - so I prefer to be more aggressive when cleaning up things.

Issue #19, opened 2025-07-23
[Task] Add a proper LICENSE file


gilcu3:
### Background

The repo needs a proper LICENSE

### User Story

OSS is expected to have a proper a license

### Acceptance Criteria

A LICENSE file is added to the repo

gilcu3:
Makes sense, adding the copyright notice should be enough. I will propose the change in #15 so that once that is merged this issue can be closed.

Issue #5, opened 2025-07-10
Rework Readme and Documentation of the implementation


SimonRastikian:
We now have an outdated documentation of this implementation and unmatching readme files.
To add/change:
- Generic DKG
- Generic types
- Eddsa
- Robust ECDSA passive and active
- crypto primitives in src/crypto folder
- Echo Broadcast
- Serialization and libraries used
- Code Comments that include step 1 step 2 etc.. (usually in triples folder, e.g.  https://github.com/near/threshold-signatures/pull/22/files#r2251477260)

SimonRastikian:
Issue for readme.md #248

SimonRastikian:
Issue for dkg #99

SimonRastikian:
PR for ckd doc #17

PR #21, opened 2025-08-01
License, Readme and doc slight modifications

SimonRastikian:
This task is inherited from the simon/passive-robust-ecdsa branch (PR #15)  and tends to improve upon the documentation of the whole repo.

Linked to issue #5 and #19.

More should be done in this regard but so I set the PR as draft unless I am advised to keep it as it is.


LICENSE
```diff
@@ -1,4 +1,5 @@
 Copyright (c) 2023 Lúcás C. Meier
+Copyright (c) 2025 NEAR One Limited
```
kuksag, line 2: I'm not sure how exactly licenses work, but can we actually take original repo, add a couple of changes and claim copyrights for that?

SimonRastikian: I have no idea either. @netrome should know more about these things.
This change was proposed by @gilcu3  and validated by @netrome

gilcu3: @kuksag I think that's the case when doing noticeable modifications, as is the case here. But I am also not an expert. I would suggest though adding `MIT` in the first line, I think it is not required but helps people looking at the file to identify which license is it.

README.md
```diff
@@ -1,2 +1,71 @@
...
+
+* Both implemented ECDSA and Ed25519 schemes do not currently provide **Robustness** i.e. recovery in case a participants drops out during presigning/signing.
+
+* Our ECDSA signing scheme outsources the message hash to the function caller (i.e. expects a hashed message as input and does not internally hash the input). However, our EdDSA implementation does not outsource the message hashing instead internally performs the message hash. This distinction is an artifact of the multiple different verifiers implemented in the wild where some might perform a "double hashing" and others not.
+(See \[[PoeRas24](https://link.springer.com/chapter/10.1007/978-3-031-57718-5_10)\] for an in-depth security study of ECDSA with outsourced hashing).
+
+* This implementation allows arbitrary number of parties and thresholds as long as the latter verifies some basic requirements (see the documentation). However, it is worth mentioning that the ECDSA scheme scales non-efficiently with the number of participants (See benchmarks).
```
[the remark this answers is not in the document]
gilcu3: line 45:
```
This implementation allows arbitrary number of parties and thresholds as long as the latter verifies some basic requirements (see the documentation).
```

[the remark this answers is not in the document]
gilcu3: Then we should add a link, markdown supports links to other files

PR #24, opened 2025-08-04
chore: use thiserror crate

kevindeforth:
resolves https://github.com/near/threshold-signatures/issues/43

### DIFF
--- a/src/protocol/errors.rs
+++ b/src/protocol/errors.rs
@@ -1,9 +1,8 @@
+use crate::protocol::Participant;
 use std::error;
 use thiserror::Error;
 
-use super::Participant;
-
-#[derive(Debug, Error)]
+#[derive(Debug, Error, PartialEq, Eq, Clone)]
 pub enum ProtocolError {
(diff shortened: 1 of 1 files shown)


src/protocol/errors.rs
SimonRastikian, line 1: Does it make sense to put some of these errors under "ProtocolError" even though they might not be directly linked to the protocol but to the underlying primitives, e.g. in polynomial.rs

src/protocol/errors.rs
```diff
@@ -0,0 +1,72 @@
+use std::error;
+use thiserror::Error;
+
+use super::Participant;
+
+#[derive(Debug, Error)]
```
netrome, line 6: Nit: Since this type is public it's nice to eagerly implement common traits for it as per https://rust-lang.github.io/api-guidelines/interoperability.html. Especially `PartialEq` and `Clone` tends to be nice to have in place when you need them for e.g. tests or whatnot.

netrome (approved):
Nice stuff! Just a few nits from my side.

PR #28, opened 2025-08-07
Simon/confidential key derivation

### LINKED ISSUE #29: [Task] Implement cryptography of the CKD
### Background

The documentation is treated in Task #1.

### User Story

This functionality will be used along with TEEs and will allow apps running on TEE to deterministically extract the same key as long as they are able to authenticate themselves.

### Acceptance Criteria

Enough test cases.
Peer review

_No response_

SimonRastikian:
CKD cryptography functionality documented in PR #17.
Closes #29 

May need additional security tests - to be answered by reviewers.


src/confidential_key_derivation/ckd.rs
```diff
@@ -0,0 +1,278 @@
...
+        return Err(InitializationError::BadParameters(format!(
+            "participant list must contain {me:?}"
+        )));
+    };
+    // ensure the coordinator is a participant
+    if !participants.contains(coordinator) {
+        return Err(InitializationError::BadParameters(format!(
+            "participant list must contain coordinator {coordinator:?}"
+        )));
+    };
```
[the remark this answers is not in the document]
SimonRastikian: I do not want to change any other file because it would create several conflicts with code-refactor and tests-generalization branches

src/confidential_key_derivation/mod.rs
```diff
@@ -0,0 +1,50 @@
...
+    pub public_key: VerifyingKey,
+}
+
+pub(crate) type CoefficientCommitment = frost_core::keys::CoefficientCommitment<Secp256K1Sha256>;
+pub(crate) type Element = frost_core::Element<Secp256K1Sha256>;
+pub(crate) type Scalar = frost_core::Scalar<Secp256K1Sha256>;
+
+/// The output of the confidential key derivation protocol when run by the coordinator
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub struct CKDCommitments {
```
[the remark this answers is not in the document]
SimonRastikian: I chose CKDCoordinatorOutput in [6aec5cb](https://github.com/near/threshold-signatures/pull/28/commits/6aec5cb9957a545d3afbb72786e643db275d8fc3)

src/confidential_key_derivation/ckd.rs
kevindeforth, line 1: What's the reason for not having them in the protocol folder?

src/confidential_key_derivation/ckd.rs
```diff
@@ -0,0 +1,278 @@
...
+
+            // choose a coordinator at random
+            let index = OsRng.next_u32() % participants.len() as u32;
+            let coordinator = participants[index as usize];
+
+            let mut protocols: Vec<(Participant, Box<dyn Protocol<Output = CKDOutput>>)> =
+                Vec::with_capacity(participants.len());
+
+            for p in &participants {
+                let share = f.eval_at_participant(*p);
```
kevindeforth, line 238: I get a compilation error here. Could it be that the branch is out of sync with the refactor branch?

```bash
help: there is a method `eval_on_zero` with a similar name
    |
215 -             let msk = f.eval_at_zero().0;
215 +             let msk = f.eval_on_zero().0;
```

src/confidential_key_derivation/ckd.rs
```diff
@@ -0,0 +1,278 @@
...
+                .next()
+                .map(|(_, c)| c.unwrap())
+                .expect("Expected exactly one Some(CKDCoordinatorOutput)");
+            assert!(
+                some_iter.next().is_none(),
+                "More than one Some(CKDCoordinatorOutput)"
+            );
+
+            // compute msk . H(app_id)
+            let confidential_key = ckd.unmask(app_sk);
```
kevindeforth, line 268: This can be slightly simplified
```rust
            // test one single some for the coordinator
            let ckd: Vec<CKDCoordinatorOutput> = result
                .into_iter()
                .filter(|(_, ckd)| ckd.is_some())
                .map(|(_, c)| c.unwrap())
                .collect();
            assert!(ckd.len() == 1);
            // compute msk . H(app_id)
            let confidential_key = ckd[0].unmask(app_sk);
```

src/confidential_key_derivation/ckd.rs
SimonRastikian: Yes, it is out of sync. Not sure what to do because if I switch back to eval_on the CI fails

gilcu3: That's very weird, what's the difference between CI and what you are running locally?

SimonRastikian: I tried syncing it back. Pull and try it now..

SimonRastikian: The CI was trying to merge into the code-refactor branch and failing due to changes there. I tried solving it for the CI by changing the function but this seems to have created local problems. I thus had to rebase to make it work on both sides...

src/confidential_key_derivation/ckd.rs
```diff
@@ -0,0 +1,278 @@
...
+        private_share,
+        app_id,
+        app_pk,
+    );
+    Ok(make_protocol(comms, fut))
+}
+
+/// Depending on whether the current participant is a coordinator or not,
+/// runs the ckd protocol as either a participant or a coordinator.
+async fn fut_wrapper(
```
netrome, line 155: I see this is consistent with existing code but I find this name very hard to comprehend. I'd expect `fut_wrapper` to be a function that contains some type of wrapper over a future, but this just calls the protocol methods. Based on the comment `run_ckd_protocol` would be a more informative name imo.

src/confidential_key_derivation/ckd.rs
```diff
@@ -0,0 +1,278 @@
...
+    chan: SharedChannel,
+    coordinator: Participant,
+    me: Participant,
+    participants: ParticipantList,
+    private_share: SigningShare,
+    app_id: Vec<u8>,
```
netrome, line 161: It would be nice if we could define a dedicated type for the app ID, along the lines of:

```
#[derive(Clone, Copy, PartialEq, Eq ...)]
struct AppId(Vec<u8>);

impl AsRef<[u8]> for AppId;
etc...
```

A nice optimization would also be to use Arc<[u8]> instead of Vec as per https://www.youtube.com/watch?v=A4cKi7PTJSs. This would save us some clones, but isn't really important.

netrome: (Can be done as a follow-up)

src/confidential_key_derivation/ckd.rs
```diff
@@ -0,0 +1,278 @@
...
+        let app_id = b"Hello Near!";
+        let pt2 = hash2curve(app_id).unwrap();
+        assert!(pt1 != pt2);
+    }
+
+    #[test]
+    fn test_ckd() -> Result<(), Box<dyn Error>> {
+        let threshold = 3;
+
+        for i in 0..100 {
```
netrome, line 207: Why do we do this 100 times?

src/confidential_key_derivation/ckd.rs
```diff
@@ -0,0 +1,278 @@
...
+}
+
+async fn do_ckd_participant(
+    mut chan: SharedChannel,
+    participants: ParticipantList,
+    coordinator: Participant,
+    me: Participant,
+    private_share: SigningShare,
+    app_id: &[u8],
+    app_pk: CoefficientCommitment,
```
netrome, line 35: Why is this a CoefficientCommitment? Shouldn't this just be a Secp256k1 pubkey? as per the design: https://github.com/near/threshold-signatures/blob/main/docs/confidential_key_derivation.md

> - $`(a,A[=a \cdot G])`$: fresh key pair generated by *app*

I undestand a `CoefficientCommitment` effectively also is a curve point but it feels like a weird type to use here.

src/confidential_key_derivation/ckd.rs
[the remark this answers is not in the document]
SimonRastikian: Would you be fine if I mark this as resolved?

src/confidential_key_derivation/ckd.rs
netrome: Using and exposing newtypes is pretty idiomatic, so I don't think this would discourage anyone from using the library. Though one way to make it extra ergonomic is to have public functions take `impl Into<AppId>` rather than `AppId` directly, in which case the user could still pass a `Vec` if they wanted.

Anyway, not a blocker for this PR. Adding more type safety is a good follow-up.

Issue #9, opened 2025-07-10
[Task] Testing Framework


SimonRastikian:
### Background

The current tests cover majorly the correctness of  the implemented schemes and less the security.
The idea is to build a framework that is able to test many correctness properties (incl. robustness, resharing and so on) but at the same time test some potential malicious behaviors.

### User Story

The aim is to augment the security assurances by having more diverse test cases. 

### Acceptance Criteria

Multiple reviews and passing test cases.
Easy use of the framework.

PR #37, opened 2025-08-15
Simon/unify-tests-interfaces 

SimonRastikian:
This PR is meant to reduce some redundancy in the tests cases and to start allowing generalization of different interfaces.
This tackles a first part of Task https://github.com/near/threshold-signatures/issues/9

The PR tackles two things:
* Simplification of the creation of participants in the tests
* Unification of running DKG functions

A file test.rs was created. It is meant to contain generic testing functions hopefully serving the multiple schemes we have.
Thanks to this file a lot of code is kicked away and replaced with one-liners


gilcu3:
@SimonRastikian when you get some time I need to know which commits should I keep here when doing rebase. After the parent branch was rebased (or before, when doing force-push) the original commits are obscured so I cannot figure anymore what was added here.

src/ecdsa/ot_based_ecdsa/test.rs
```diff
@@ -68,18 +64,42 @@ pub fn run_presign(
...
+    assert_public_key_invariant(&key_packages);
+    let (pub0, shares0) = deal(&mut OsRng, &participants, threshold).unwrap();
+    let (pub1, shares1) = deal(&mut OsRng, &participants, threshold).unwrap();
+
+    // Presign
+    let mut presign_result = run_presign(key_packages, shares0, shares1, &pub0, &pub1, threshold);
+    presign_result.sort_by_key(|(p, _)| *p);
+
+    let msg = b"hello world";
+    run_sign(
```
gilcu3, line 87: Maybe add an explanatory comment here mentioning that this also checks the signature validity

src/test.rs
```diff
@@ -0,0 +1,144 @@
...
+            &new_participants,
+            new_threshold,
+            *p,
+        )?;
+        protocols.push((*p, Box::new(protocol)));
+    }
+
+    let result = run_protocol(protocols)?;
+    Ok(result)
+}
```
gilcu3, line 130: Afaik this test is checking the case when resharing is done with a new participant. Can we also check the other limits? For example when the number of new participants is the maximum possible, or when some participants leave? My point is we should cover all edge cases.

gilcu3 (approved):
Great job! 
- I added mainly a comment about covering edges cases for resharing

src/ecdsa/ot_based_ecdsa/test.rs
SimonRastikian: Done! All this is going to be done away with in the upcoming PR so that one has a single signing test function

src/test.rs
SimonRastikian: This is not true. This function is called when a reshare should happen. Look for example at eddsa/sign.rs where run_reshare is called in both test_reshare_sign_less_participants, and test_reshare_sign_more_participants.

What do you mean when the number of participants is the maximum possible? This would be 2^128 if not more.

SimonRastikian: Ok I will open an issue for adding these test cases, and add an extra test case for checking schemes robustness. Maybe also add some cases for malicious signers.

SimonRastikian:
I made a mistake force pushing a wrong rebase. I had to revert the force push by restoring the commits. This is why it seems as if the commits b3bc64e to da8f78e have not been verified yet. But all good. I resolved a small conflict in generate_participants.

gilcu3:
As agreed in https://nearone.slack.com/archives/C0912BTG51T/p1756374506678839?thread_ts=1756368856.422359&cid=C0912BTG51T I will rebase this PR on main.

Issue #49, opened 2025-08-28
Add extra error types


SimonRastikian:
InitializationError contains only BadParameter.
One can add extra types that tell exactly what is wrong with the parameter, simplifying the code review.

PR #52, opened 2025-09-01
refactor(#49): make InitializationError more specific

0xsecaas:
This commit refactors the `InitializationError` enum to be more specific about the error conditions. Instead of a generic `BadParameters(String)` error, we now have more specific errors like `DuplicateParticipants`, `NotEnoughParticipants`, `ThresholdTooSmall`, `ThresholdTooLarge`, and `InvalidParticipantIndex`.

This makes the code easier to debug and review, as the error type now carries more information about what went wrong.

This commit also adds a unit test to ensure that the correct error is returned when duplicate participants are provided.

### DIFF
--- a/src/confidential_key_derivation/protocol.rs
+++ b/src/confidential_key_derivation/protocol.rs
@@ -129,9 +129,10 @@ pub fn ckd(
     };
     // ensure the coordinator is a participant
     if !participants.contains(coordinator) {
-        return Err(InitializationError::BadParameters(format!(
-            "participant list must contain coordinator {coordinator:?}"
-        )));
+        return Err(InitializationError::MissingParticipant {
+            role: "coordinator",
+            participant: coordinator,
+        });
     };
 
     let comms = Comms::new();
--- a/src/confidential_key_derivation/protocol.rs
+++ b/src/confidential_key_derivation/protocol.rs
@@ -111,17 +111,14 @@ pub fn ckd(
 ) -> Result<impl Protocol<Output = CKDOutput>, InitializationError> {
     // not enough participants
     if participants.len() < 2 {
-        return Err(InitializationError::BadParameters(format!(
-            "participant count cannot be < 2, found: {}",
-            participants.len()
-        )));
+        return Err(InitializationError::NotEnoughParticipants {
+            participants: participants.len() as u32,
+        });
     };
 
     // kick out duplicates
     let Some(participants) = ParticipantList::new(participants) else {
-        return Err(InitializationError::BadParameters(
-            "Participants list contains duplicates".to_string(),
-        ));
+        return Err(InitializationError::DuplicateParticipants);
     };
 
     // ensure my presence in the participant list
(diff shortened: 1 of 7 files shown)


src/protocol/errors.rs
```diff
@@ -71,8 +71,20 @@ impl From<Box<dyn error::Error + Send + Sync>> for ProtocolError {
...
-    #[error("bad parameters: {0}")]
-    BadParameters(String),
+    #[error("bad parameter: {0}")]
+    BadParameter(String),
+    #[error("participant list cannot contain duplicates")]
+    DuplicateParticipants,
+    #[error(
+        "threshold {threshold} cannot be larger than the number of participants {participants}"
+    )]
+    NotEnoughParticipants { threshold: u32, participants: u32 },
```
gilcu3, line 83: I see that you tried to generalize the error of not having enough participants, and not having enough participants for a given threshold. I would rather keep the separate
```suggestion
    #[error( "participant count cannot be < 2, found: {participants}" )]
    NotEnoughParticipants { participants: u32 },
    #[error(
        "not enough intersecting old/new participants ({participants}) to reconstruct private key for resharing with threshold bigger than old threshold ({threshold})"
    )]
    NotEnoughParticipantsForThreshold { threshold: u32, participants: u32 },
```

src/generic_dkg.rs
```diff
@@ -597,52 +597,41 @@ pub(crate) fn reshare_assertions<C: Ciphersuite>(
...
+    let old_participants =
+        ParticipantList::new(old_participants).ok_or(InitializationError::DuplicateParticipants)?;
 
     if old_participants.intersection(&participants).len() < old_threshold {
-        return Err(InitializationError::BadParameters(
-            format!("not enough intersecting old/new participants {:?} to reconstruct private key for resharing with threshold bigger than old threshold {}",
-            old_participants.intersection(&participants),
-            old_threshold).to_string(),
-        ));
+        return Err(InitializationError::NotEnoughParticipants {
```
gilcu3, line 625:
```suggestion
        return Err(InitializationError::NotEnoughParticipantsForThreshold {
```

src/generic_dkg.rs
```diff
@@ -597,52 +597,41 @@ pub(crate) fn reshare_assertions<C: Ciphersuite>(
     old_participants: &[Participant],
 ) -> Result<(ParticipantList, ParticipantList), InitializationError> {
     if participants.len() < 2 {
-        return Err(InitializationError::BadParameters(format!(
-            "participant count cannot be < 2, found: {}",
-            participants.len()
-        )));
+        return Err(InitializationError::NotEnoughParticipants {
+            threshold: 2,
```
gilcu3, line 601:
```suggestion
```

src/eddsa/sign.rs
```diff
@@ -212,26 +212,24 @@ pub fn sign(
     message: Vec<u8>,
 ) -> Result<impl Protocol<Output = SignatureOutput>, InitializationError> {
     if participants.len() < 2 {
-        return Err(InitializationError::BadParameters(format!(
-            "participant count cannot be < 2, found: {}",
-            participants.len()
-        )));
+        return Err(InitializationError::NotEnoughParticipants {
+            threshold: 2,
```
gilcu3, line 216:
```suggestion
```

gilcu3 (changes requested):
Thank you very much!
Mostly minor changes requested, will check again once CI works and it is rebased on main.

src/confidential_key_derivation/protocol.rs
```diff
@@ -271,4 +268,33 @@ mod test {
...
+            me,
+            private_share,
+            app_id,
+            app_pk,
+        );
+        match result {
+            Ok(_) => panic!("Expected an error, but got Ok"),
+            Err(err) => assert_eq!(err, InitializationError::DuplicateParticipants),
+        }
+    }
```
SimonRastikian, line 299: Nit: Replace this by -> Result<(); dyn Box> in the function signature then call ckd and add ? operator at the end.

SimonRastikian:
All good!

Issue #47, opened 2025-08-25
Adding more type safety


SimonRastikian:
Confidential key derivation could benefit from a type safety namely:
```
#[derive(Clone, Copy, PartialEq, Eq ...)]
struct AppId(Vec<u8>);

impl AsRef<[u8]> for AppId;
etc...
```
or using Arc<[u8]> 

Reference the following comment:
https://github.com/near/threshold-signatures/pull/28#discussion_r2288159908

PR #64, opened 2025-09-06
feat(ckd): Improve AppId type safety and optimize with Arc #47 

0xsecaas:
- Introduces `AppId` newtype for enhanced type safety and clarity in CKD.
- Optimizes `AppId` to use `Arc<[u8]>` for cheaper cloning and shared ownership,

Refs #47


src/confidential_key_derivation/mod.rs
```diff
@@ -1,3 +1,44 @@
...
+//! the application identifier to the key derivation service.
+//!
+//! The protocol is based on a combination of Oblivious Transfer (OT) and Diffie-Hellman key exchange.
+//!
+//! For more details, refer to the `confidential_key_derivation.md` document in the `docs` folder.
+
+use std::sync::Arc;
+
+#[derive(Clone, PartialEq, Eq, Debug, Hash, PartialOrd, Ord)]
+pub struct AppId(Arc<[u8]>);
```
netrome, line 14: While this is a great start, a few additions would make this more ergonomic and could be useful downstream:

1. Add serialization implementations (serde and borsh).
2. Add a display method returning a hex encoded string.

netrome (dismissed):
Thank you for doing this. The only blocker from my side is that the `From` implementation creates a redundant imtermediate Box which can be skipped to make the code both simpler and more effective.

Beyond that I added a bunch of suggestions, but feel free to skip any of them. None are strictly necessary and all of them can be added later.

PR #79, opened 2025-09-13
fix(test): consider debug errors #71

### LINKED ISSUE #71: [Task] Safe and performant test coverage
### Background

Relying only on `--release` risks missing _debug-mode_ bugs (e.g., _overflow_,  _panics_, _debug assertions_). Plain `cargo test` is slow (~1 min), while `cargo nextest` is fast (<1s). `cargo test --profile test-release` is a middle ground: it catches debug-mode issues quickly (<1s) and keeps development efficient. 

Goal: ensure safety, performance, and convenience.

### User Story

As a developer, I need tests that catch safety issues and run super fast, so I can trust correctness while keeping development efficient.

### Acceptance Criteria

[ ] Add `test-release` profile to [`cargo.toml`](https://github.com/near/threshold-signatures/blob/main/Cargo.toml):

```toml
[profile.test-release]
inherits = "release"
overflow-checks = true
debug = true
```

[ ] update [`Makefile`](https://github.com/near/threshold-signatures/pull/70/files#diff-9375fd04332c86472d7be397ef09428cb86babd8826880a5835bd1d1c1bdbc08) to run `cargo test --profile test-release` instead of `nextest` for custom-profile support.
[ ]  [CI](https://github.com/near/threshold-signatures/blob/main/.github/workflows/ci.yml) runs both: 
- `cargo test` (debug)
- `cargo nextest run --release`

_No response_

0xsecaas:
Closes #71

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -21,7 +21,7 @@ args = [
   "--profile",
   "test-release",
   "--all-features",
-  "all-targets",
+  "--all-targets",
   "--locked",
 ]
 


Makefile.toml
```diff
@@ -16,7 +16,7 @@ args = ["fmt", "--", "--check"]
 
 [tasks.test]
 command = "cargo"
-args = ["nextest", "run", "--release", "--all-features", "--all-targets", "--locked"]
+args = ["test", "--profile", "test-release"]
```
gilcu3, line 19: don't we still need all the flags?

gilcu3 (approved):
Thank you! Just a minor question/change.

.github/workflows/ci.yml
```diff
@@ -46,6 +46,9 @@ jobs:
       - name: Run cargo-nextest
         run: cargo nextest run --release --locked --all-features
 
+      - name: Run cargo test debug
+        run: cargo test
```
gilcu3, line 50: and here as well, we should probably use `--locked --all-features --all-targets`

Makefile.toml
0xsecaas: Good call, thanks!

Makefile.toml
```diff
@@ -16,7 +16,14 @@ args = ["fmt", "--", "--check"]
 
 [tasks.test]
 command = "cargo"
-args = ["nextest", "run", "--release", "--all-features", "--all-targets", "--locked"]
+args = [
+  "test",
+  "--profile",
+  "test-release",
+  "--all-features",
+  "all-targets",
```
gilcu3, line 24: I think you missed a couple of hyphens there

PR #80, opened 2025-09-13
fix(security): use constant-time cryptographic operations to mitigate side-channel attacks

0xsecaas:
Utilize the _constant-time_ checks across various components, mitigating potential timing side-channel attacks.

  Key changes include:
   - Applied ConstantTimeEq to `Commitment` and `HashOutput`.
   - Enhanced `batch_compute_lagrange_coefficients` with constant-time early exit (**3%** improvement based on current bencmarks).
   - Ensured `dlogeq` proof and verification perform identity checks in constant time.
   - Added _constant-time_ checks for identity elements and zero values in robust ECDSA presigning and signing.

> New tests verify the functional correctness of these constant-time implementations.

### DIFF
--- a/src/crypto/polynomials.rs
+++ b/src/crypto/polynomials.rs
@@ -483,22 +483,22 @@ where
     // If x exactly equals some x_i, return Kronecker delta vector
     // This is done in constant time by iterating through all elements
     // and accumulating a Choice without short-circuiting.
-    let mut k_index_val = CtOption::new(0u32, Choice::from(0u8)); // Initialize as CtOption::none() effectively
+    let mut kronecker_index = CtOption::new(0u32, Choice::from(0u8)); // Initialize as CtOption::none() effectively
 
     for (i, p) in points_set.iter().enumerate() {
         let is_equal = p.ct_eq(x);
         // If is_equal is true, select 'i', otherwise keep the current k_index_val
-        k_index_val = CtOption::conditional_select(
-            &k_index_val,
+        kronecker_index = CtOption::conditional_select(
+            &kronecker_index,
             &CtOption::new(i as u32, is_equal),
             is_equal,
         );
     }
 
-    if k_index_val.is_some().into() {
-        let k_index = k_index_val.unwrap() as usize;
+    if kronecker_index.is_some().into() {
+        let kronecker_index_value = kronecker_index.unwrap() as usize;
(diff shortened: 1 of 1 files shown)


src/crypto/polynomials.rs
```diff
@@ -471,9 +481,24 @@ pub fn batch_compute_lagrange_coefficients<C: Ciphersuite>(
...
+
+    for (i, p) in points_set.iter().enumerate() {
+        let is_equal = p.ct_eq(x);
+        // If is_equal is true, select 'i', otherwise keep the current k_index_val
+        k_index_val = CtOption::conditional_select(
+            &k_index_val,
+            &CtOption::new(i as u32, is_equal),
+            is_equal,
+        );
+    }
```
SimonRastikian, line 496: Why would this work if you are changing k_index_val in each round? Is this not equivalent to do it only for the last round?

src/crypto/polynomials.rs
```diff
@@ -471,9 +481,24 @@ pub fn batch_compute_lagrange_coefficients<C: Ciphersuite>(
...
+
+    if k_index_val.is_some().into() {
+        let k_index = k_index_val.unwrap() as usize;
         let mut coeffs = vec![SerializableScalar(<C::Group as Group>::Field::zero()); n];
-        coeffs[k] = SerializableScalar(<C::Group as Group>::Field::one());
+        coeffs[k_index] = SerializableScalar(<C::Group as Group>::Field::one());
```
SimonRastikian, line 501: I would call the variable kronecker instead of k as the latter is used a lot in ecdsa as a private ephemeral share

src/crypto/polynomials.rs
0xsecaas: Not quite, we update `k_index_val` every round so we don’t miss the index where `p` matches `x`.

Because the update only happens when `is_equal` is true, the loop leaves `k_index_val` set to the index of the match once it’s done.


If `points_set` is unique, there can only be one match, so this works as expected. If there are multiple matches, we’d end up with the last matching index.

SimonRastikian: Oh yes, you're right otherwise we keep the k_index_val as it is.

SimonRastikian (approved):
Approved! Nice work :+1:

SimonRastikian:
Uses constant time comparison and assignment in the crypto crate.

PR #84, opened 2025-09-17
fix(#49): use proper error, and add the hashing one

0xsecaas:
Continues  #49


gilcu3:
If this PR is included in https://github.com/near/threshold-signatures/pull/85 feel free to close it, I am merging that one

Issue #59, opened 2025-09-05
Add the rerandomization and key derivation functions in this repo


SimonRastikian:
Currently the Rerandomization and Key derivation [GS21] are done on the MPC side. We would like to have them in this repo instead to encapsulate all cryptography functions here.
This issue is only related so far with the ot_based_ecdsa scheme.

Accpetance Criteria:
1) Peer reviewed code.
2) Easy to use API
3) Tested rerandomization signing
4) Implementation forces the library user to rerandomize the presignature.

PR #86, opened 2025-09-17
Simon/rerandomization api ot ecdsa

SimonRastikian:
Adding rerandomization and key derivation of the presignature API is added to this repo in order to improve on the security of the scheme following [GS21]. The structures force the library user to rerandomize their presignatures.
This requires modifying (relaxing) the framework of MPC repo.

This PR is tightly related to https://github.com/near/threshold-signatures/issues/59.


src/ecdsa/mod.rs
```diff
@@ -68,24 +72,120 @@ impl Signature {
...
+        // concatenate all the bytes
+        let mut concatenation = Vec::new();
+        concatenation.extend_from_slice(encoded_pk);
+        concatenation.extend_from_slice(encoded_msg_hash);
+        concatenation.extend_from_slice(encoded_big_r);
+        // Append each ParticipantId's
+        for participant in self.participants.participants() {
+            concatenation.extend_from_slice(&participant.bytes());
+        }
+
```
gilcu3, line 140: For correctness, this assumes that all participants get the same order right?

src/ecdsa/mod.rs
```diff
@@ -68,24 +72,120 @@ impl Signature {
...
+            // append an extra 0 at the end of the concatenation everytime delta hits zero
+            concatenation.extend_from_slice(&[0u8, 1]);
+            hk.expand(&concatenation, &mut okm).unwrap();
+
+            // derive the randomness delta
+            delta = Scalar::from_repr(okm.into()).unwrap_or(
+                // if delta falls outside the field
+                // probability is negligible: in the order of 1/2^224
+                Scalar::ZERO,
+            )
```
gilcu3, line 162: Just to confirm, afaik the computation in this function uses public information, so we don't need to worry about attackers leaking information from this "random sample with abort"

src/ecdsa/mod.rs
```diff
@@ -68,24 +72,120 @@ impl Signature {
...
+        // If the randomness created is 0 then we want to generate a new randomness
+        while bool::from(delta.is_zero()) {
+            // Generate randomization out of HKDF(entropy, pk, msg_hash, big_r, participants, nonce)
+            // where entropy is a public but unpredictable random string
+            // the nonce is a succession of appended ones of growing length depending on the number of times
+            // we enter into this loop
+            let mut okm = [0u8; 32];
+
+            // append an extra 0 at the end of the concatenation everytime delta hits zero
+            concatenation.extend_from_slice(&[0u8, 1]);
```
gilcu3, line 154: crypto nit: I would put this line later in the loop, so that `concatenation` does not get the extra 0 when we succeed to obtain `delta` in the first round of the loop

src/ecdsa/mod.rs
```diff
@@ -68,24 +72,120 @@ impl Signature {
...
+
+        // concatenate all the bytes
+        let mut concatenation = Vec::new();
+        concatenation.extend_from_slice(encoded_pk);
+        concatenation.extend_from_slice(encoded_msg_hash);
+        concatenation.extend_from_slice(encoded_big_r);
+        // Append each ParticipantId's
+        for participant in self.participants.participants() {
+            concatenation.extend_from_slice(&participant.bytes());
+        }
```
gilcu3, line 139: Are you certain concatenating this way is fine? I am curious about the reasons, like why not adding a separator.

src/ecdsa/mod.rs
```diff
@@ -141,4 +241,113 @@ mod test_verify {
...
+        let participants = generate_participants_with_random_ids(num_participants);
+        let participants = ParticipantList::new(&participants).unwrap();
+
+        let mut entropy: [u8; 32] = [0u8; 32];
+        OsRng.fill_bytes(&mut entropy);
+
+        let args = RerandomizationArguments::new(&pk, &msg_hash, &big_r, &participants, entropy);
+        let delta = args.derive_randomness();
+        (pk, big_r, msg_hash, participants, entropy, delta)
+    }
```
gilcu3, line 272: Now that we can generate randomness for tests using `MockCryptoRng`, we can have a test with exactly this contents, but that at the end checks for fixed values. Another possibility is to receive the RNG as a parameter here (`rng: &mut impl CryptoRngCore`), so that the other test I am imagining becomes a "one liner"

src/ecdsa/mod.rs
```diff
@@ -141,4 +241,113 @@ mod test_verify {
...
+
+    // Test that with different order of participants, the randomness is the same.
+    #[test]
+    fn test_same_randomness() {
+        let num_participants = 10;
+        let (pk, big_r, msg_hash, participants, entropy, delta) =
+            compute_random_outputs(num_participants);
+
+        let mut rng = rand::rng();
```
gilcu3, line 344: in a PR still in review https://github.com/near/threshold-signatures/pull/78 I removed the dependency on `rand`. So, if possible, please try to use `rand_core` here.

src/ecdsa/ot_based_ecdsa/mod.rs
```diff
@@ -16,23 +40,60 @@ pub struct PresignOutput {
...
-pub mod triples;
+impl RerandomizedPresignOutput {
+    pub fn new(
+        presignature: &PresignOutput,
+        tweak: &Tweak,
+        args: &RerandomizationArguments,
+    ) -> Result<Self, ProtocolError> {
+        if presignature.big_r != *args.big_r {
+            return Err(ProtocolError::IncompatibleRerandomizationInputs);
+        }
```
gilcu3, line 64: I find this very weird. Why do we need that parameter passed twice as part of the input?

src/ecdsa/ot_based_ecdsa/test.rs
```diff
@@ -35,6 +51,57 @@ pub fn run_sign(
     )
 }
 
+type SigWithRerand = (Tweak, Vec<(Participant, Signature)>);
```
gilcu3, line 54: I learn recently that it is better to avoid this
```suggestion
type SignatureWithRerandomization = (Tweak, Vec<(Participant, Signature)>);
```

gilcu3 (changes requested):
Thank you, I think this is almost done!

Left a few suggestions, and as many questions (as usual)

src/ecdsa/ot_based_ecdsa/mod.rs
SimonRastikian: We do not. This is an artefact of the different structures both containing the same item big_r.

src/ecdsa/mod.rs
SimonRastikian: All of these items are of fixed size so should be fine

src/ecdsa/ot_based_ecdsa/test.rs
SimonRastikian: This goes away entirely with #93 I will not modify it not to cause unecessary conflicts

src/lib.rs
```diff
@@ -116,13 +154,3 @@ where
...
-
-// Libraries calls
-use crypto::ciphersuite::Ciphersuite;
-use frost_core::{keys::SigningShare, VerifyingKey};
-use serde::{Deserialize, Serialize};
-use std::marker::Send;
-
-use crate::generic_dkg::*;
-use crate::protocol::internal::{make_protocol, Comms};
-use crate::protocol::{errors::InitializationError, Participant, Protocol};
```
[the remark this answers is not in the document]
SimonRastikian: I got confused then. I specifically looked at one of the comments in a previous PR saying to put these in the beginning. I might have confused things.

src/ecdsa/ot_based_ecdsa/mod.rs
```diff
@@ -1,6 +1,30 @@
-use crate::ecdsa::{AffinePoint, KeygenOutput, Scalar};
+pub mod presign;
+pub mod sign;
+pub mod triples;
+
+#[cfg(test)]
+mod test;
```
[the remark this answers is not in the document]
SimonRastikian: I am very very confused now https://github.com/near/threshold-signatures/pull/22#discussion_r2251578469
@netrome ?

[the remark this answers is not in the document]
netrome: Yeah new module declarations should be at the top imo. Ideally before imports. I'd be curious to see if I or anyone else has claimed otherwise.

src/lib.rs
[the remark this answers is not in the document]
netrome: I don't think we ever agreed to the "imports should be at the bottom" convention. Though I did use this convention in my http disaster recovery PoC https://github.com/near/mpc/pull/921 but that's because the code was never intended for merging, and because it reads better imo.

But I apologize if I have caused any confusion about this. I don't know any other developer that promotes this convention than I, and so far it does not make sense to use it in production code due to how unfamiliar people are with the idea of putting imports on the bottom.

src/ecdsa/ot_based_ecdsa/mod.rs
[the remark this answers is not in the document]
gilcu3: in the big code refactor https://github.com/near/threshold-signatures/blame/583e0544fc1316a2b53416058841a5b9b386501b/src/ecdsa/ot_based_ecdsa/mod.rs#L35 they were put at the bottom. Was that a mistake? Or simply that it was there before?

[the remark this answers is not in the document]
netrome: Although one of my personal preferences in my own code (not code shared with others) is that imports should go on the bottom of the file.

This sounds like it stems from me promoting bottom-imports at some point. Which, to be clear, is a great way to write more readable code in my view - but it is also not widely accepted and best suited for personal side projects until we can get more buy-in from the community that this is in fact the right way to organize the code.

src/ecdsa/ot_based_ecdsa/mod.rs
gilcu3: I think we could consider the full presignature and not only `big_r` as part of the `RerandomizationArguments`, but it is true that the presignature also contains private shares that are not needed for `RerandomizationArguments`. So yeah, it might be the best solution as you deduced. Marking as resolved

src/lib.rs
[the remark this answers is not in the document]
gilcu3: > this convention in my http disaster recovery PoC https://github.com/near/mpc/pull/921

Alright! that's where I got it from then :) I am following yet another convention when coding or doing reviews. If I don't have any strong reason, I don't change the position of code. Here this was already in the bottom, now moved to the top, so I complained to follow my own convention :)

src/ecdsa/ot_based_ecdsa/mod.rs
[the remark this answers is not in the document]
netrome: > given that cargo fmt and clippy do not complain, I would say that the community is undecided on where to put them

Yeah the tooling works great with it, to my surprise. Though it makes sense. Cargo fmt won't re-order your functions and struct declarations either, so why should they reorder your imports?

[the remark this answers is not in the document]
gilcu3: same for modules https://rust-lang.github.io/rustfmt/?version=v1.8.0&search=#reorder_modules

src/lib.rs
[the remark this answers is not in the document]
SimonRastikian: I guess moral: libraries should be called in the beginning

src/lib.rs
```diff
@@ -26,6 +36,34 @@ pub struct KeygenOutput<C: Ciphersuite> {
     pub public_key: VerifyingKey<C>,
 }
 
+/// This is a necessary element to be able to derive different keys
+/// from signing shares.
+/// We do not bind the user with the way to compute the inner scalar of the tweak
+pub struct Tweak<C: Ciphersuite>(Scalar<C>);
```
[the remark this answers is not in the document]
netrome: Unfortunately not, since the `Scalar` type enforces this bound (implicitly by being a type alias). But still, good that you're keeping an eye up for these things!

src/ecdsa/ot_based_ecdsa/test.rs
```diff
@@ -35,6 +51,57 @@ pub fn run_sign(
...
+/// Runs signing by calling the generic run_sign function from crate::test
+/// This signing mimics what should happen in real world, i.e.,
+/// rerandomizing the presignatures
+pub fn run_sign_with_rerandomization(
+    participants_presign: Vec<(Participant, PresignOutput)>,
```
netrome, line 59: Taking ownership here seems redundant
```suggestion
    participants_presign: &[(Participant, PresignOutput)],
```

netrome: I see this is already and established pattern in the old function. Feel free to update or skip. Not super important.

src/ecdsa/mod.rs
```diff
@@ -68,24 +72,120 @@ impl Signature {
...
+
+    /// Derives a random string from the public key, message hash, presignature R,
+    /// set of participants and the entropy.
+    ///
+    /// Outputs a random string computed as HKDF(entropy, pk, hash, R, participants)
+    pub fn derive_randomness(&self) -> Scalar {
+        // create a string containing (pk, msg_hash, big_r, sorted(participants))
+        let pk_encoded_point = self.pk.to_encoded_point(true);
+        let encoded_pk: &[u8] = pk_encoded_point.as_bytes();
+        let encoded_msg_hash: &[u8] = &self.msg_hash.to_bytes()[..];
```
netrome, line 127: No need to slice here
```suggestion
        let encoded_msg_hash: &[u8] = &self.msg_hash.to_bytes();
```

netrome (approved):
Shallow review, looks good to me. Some nits.

src/lib.rs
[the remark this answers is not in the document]
netrome: Right correct. The point is you can't remove the bound, but it's true we don't need to enforce anything beyond `frost_core::Ciphersuite`.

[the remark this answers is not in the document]
netrome: > Another nit question, given that this is a public type, should it have also?

Good observation. How will we pass this around? It's probably safest to proactively add these traits and others as per https://rust-lang.github.io/api-guidelines/interoperability.html?highlight=eager#types-eagerly-implement-common-traits-c-common-traits

src/ecdsa/mod.rs
SimonRastikian: Done!

src/ecdsa/mod.rs
SimonRastikian: Outdated after [f3d7b6b](https://github.com/near/threshold-signatures/pull/86/commits/f3d7b6b41867deaa635e0ca6681bcde5901b5060)

src/lib.rs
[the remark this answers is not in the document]
SimonRastikian: > Good observation. How will we pass this around?
I only see such a structure used locally and not sent over the wire or cloned. I had to make it public so that another library can actually call the constructor.
Supposedly Tweak is not to be sent between participants or so. However, I did add most of those traits but had to change Scalar into SerializableScalar.

Btw, what did we conclude over the part related to doing the following? 
`pub struct Tweak<C: Ciphersuite>(...);`

src/ecdsa/mod.rs
gilcu3: Almost. There should be a test checking the functionality against fixed values, that can be obtained with a previous run using the Mock crypto

src/ecdsa/mod.rs
```diff
@@ -141,4 +246,118 @@ mod test_verify {
...
+    fn compute_random_outputs(
+        rng: &mut impl CryptoRngCore,
+        num_participants: usize,
+    ) -> (
+        AffinePoint,
+        AffinePoint,
+        [u8; 32],
+        ParticipantList,
+        [u8; 32],
+        Scalar,
```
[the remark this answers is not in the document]
SimonRastikian: Okay adapted all here [06b9610](https://github.com/near/threshold-signatures/pull/86/commits/06b9610d8eb8fe3d54768956c58249eabadee74c)

src/ecdsa/ot_based_ecdsa/sign.rs
```diff
@@ -21,7 +21,7 @@ pub fn sign(
     participants: &[Participant],
     me: Participant,
     public_key: AffinePoint,
-    presignature: PresignOutput,
+    presignature: RerandomizedPresignOutput,
```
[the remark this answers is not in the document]
SimonRastikian: I ask the question differently: Why should we not expose the functionalities? What if someone wants to perform a batching for the rerandomization by running multiple presignatures?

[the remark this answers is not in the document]
gilcu3: That's a good question. Do we want to cover such a case? I thought the point of doing re-randomization this way was to enforce a correct use of the library. Doing that while remaining minimal, implies that the functions should not be public. If you think realistically someone could use it in some other way, then having them public is fine.

gilcu3 (approved):

PR #89, opened 2025-09-17
Simon/differences robust ecdsa doc

SimonRastikian:
This PR adds a section to the documentation that declares the differences between our spec and the original scheme.
A couple of additional small typos are repaired in the previous sections.


gilcu3:
I think the rebase did not work here @SimonRastikian . If I remember correctly, this was a PR with a small addition, now it looks like it is modifying a lot of files

netrome:
Yeah seems like this should be rebased on latest main. Not sure why the robust-ecdsa-doc branch was merged into this one https://github.com/near/threshold-signatures/pull/89/commits/9d439330e532ed08d3a948003bddfe15d170e2f6

PR #93, opened 2025-09-17
Adding coordinator and Participants to Robust ECDSA scheme

SimonRastikian:
This PR aims to introduce asymmetry in the signing by having a coordinator and participants for the robust ECDSA scheme.

We added in the tests.rs files a generic function for this functionality.
Due to now the existence of two testing functions that runs the signing protocol, we would be happy to have an upcoming PR that transform the online phase of the OT-based ECDSA Scheme.

### DIFF
--- a/src/ecdsa/mod.rs
+++ b/src/ecdsa/mod.rs
@@ -72,6 +72,9 @@ impl Signature {
     }
 }
 
+/// None for participants and Some for coordinator
+pub type SignatureOption = Option<Signature>;
+
 /// The arguments used to derive randomness used for presignature rerandomization.
 /// Presignature rerandomization has been thoroughly described in
 /// [GS21] https://eprint.iacr.org/2021/1330.pdf
--- a/src/ecdsa/robust_ecdsa/sign.rs
+++ b/src/ecdsa/robust_ecdsa/sign.rs
@@ -6,7 +6,7 @@ use subtle::ConditionallySelectable;
 use crate::{
     ecdsa::{
         robust_ecdsa::RerandomizedPresignOutput, x_coordinate, AffinePoint, Polynomial, Scalar,
-        Secp256K1Sha256, Signature,
+        Secp256K1Sha256, Signature, SignatureOption,
     },
     participants::{ParticipantCounter, ParticipantList, ParticipantMap},
     protocol::{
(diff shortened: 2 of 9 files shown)


SimonRastikian:
If we transform the OT-based ECDSA scheme into asymmetric with a Coordinator, then we can merge some of the code related to signature testing (src/test.rs) and make (again) the signature structure uniform across the schemes, i.e. Option<Signature>

src/protocol/errors.rs
```diff
@@ -52,6 +52,9 @@ pub enum ProtocolError {
     #[error("the constructed signing key is null")]
     MalformedSigningKey,
 
+    #[error("Expected exactly one output that belongs only to the coordinator")]
+    MismatchCoordinatorOutput,
+
```
gilcu3, line 57: Could this error ever happen, except in our testing functions? If yes, maybe feature gating it makes sense

src/test.rs
```diff
@@ -178,6 +181,74 @@ where
...
+/// Runs the signing algorithm for ECDSA.
+/// The scheme must be asymmetric as in: there exists a coordinator that is different than participants.
+/// Only used for unit tests.
+pub(crate) fn run_asymmetric_sign<C: Ciphersuite, PresignOutput, Signature: Clone, F>(
+    participants_presign: Vec<(Participant, PresignOutput)>,
+    coordinator: Participant,
+    public_key: frost_core::Element<C>,
+    msg_hash: frost_core::Scalar<C>,
+    sign: F,
+) -> Result<Vec<(Participant, Signature)>, Box<dyn Error>>
```
gilcu3, line 193: I am not totally sure why such a function is needed. If a test uses this function, instead of the `run_sign` that is actually used by the library users, what would be the benefit? Shouldn't tests be using the public functions?

gilcu3 (approved):
Looks good!

Left a couple of questions/suggestions.

- I noticed some names that changed here, also changed again in https://github.com/near/threshold-signatures/pull/94, would be good to settle on names before merging to main

src/ecdsa/mod.rs
```diff
@@ -72,6 +72,9 @@ impl Signature {
     }
 }
 
+/// None for participants and Some for coordinator
+pub type SignatureOption = Option<Signature>;
```
netrome, line 76: Nit: Is this a helpful type alias? I'd consider skipping it.

src/ecdsa/robust_ecdsa/sign.rs
```diff
@@ -17,13 +17,16 @@ use crate::{
 };
 type C = Secp256K1Sha256;
 
+/// Depending on whether the current participant is a coordinator or not,
+/// runs the signature protocol as either a participant or a coordinator.
 pub fn sign(
     participants: &[Participant],
+    coordinator: Participant,
```
netrome, line 24: Passing the coordinator around feels redundant if all we end up doing is comparing it with "me". Could we simplify this to?
```suggestion
    is_coordinator: bool,
```

netrome (approved):
Had a quick look. Some considerations from my side, but all in all it looks good to me.

PR #101, opened 2025-09-23
feat: implement CKD using blstrs

### LINKED ISSUE #104: Use blstrs in CKD
### Background

We recently decided to pivot our current CKD implementation to BLS.

### User Story

BLS will add receiver verifiability basically for free.

### Acceptance Criteria

CKD works with blstrs, all tests are implemented and passing

_No response_

gilcu3:
Closes #104 

Includes:
- `DKG`, which was free after implementing `Ciphersuite` trait
- `CKD`
- basic unit tests
- basic integration test to make sure all types needed are public
- docs

Does not include:
- `hash_to_field`
- thorough testing

Might use more testing. Let me know which ones would you consider appropriate, might be good to implement here before merging.

For simplicity, I used the `Projective` variants of point in curve types everywhere, instead of using `Affine` variants  as return types, which is better (smaller size). I think to avoid complicating this PR, if that's appropriate we should do it in a follow-up.

### DIFF
--- a/src/test.rs
+++ b/src/test.rs
@@ -290,6 +290,7 @@ impl RngCore for MockCryptoRng {
     }
 }
 
+// Taken from https://github.com/ZcashFoundation/frost/blob/3ffc19d8f473d5bc4e07ed41bc884bdb42d6c29f/frost-secp256k1/tests/common_traits_tests.rs#L9
 #[allow(clippy::unnecessary_literal_unwrap)]
 pub fn check_common_traits_for_type<T: Clone + Eq + PartialEq + std::fmt::Debug>(v: T) {
     // Make sure can be debug-printed. This also catches if the Debug does not


src/confidential_key_derivation/ciphersuite.rs
```diff
@@ -0,0 +1,309 @@
...
+    let base2 =
+        <<BLS12381SHA256 as frost_core::Ciphersuite>::Group as frost_core::Group>::generator()
+            .into();
+    let element2 = signature.into();
+    if blstrs::pairing(&base1, &element1) == blstrs::pairing(&element2, &base2) {
+        Ok(())
+    } else {
+        Err(frost_core::Error::InvalidSignature)
+    }
+}
```
netrome, line 232: While complex under the hood, BLS signature verification is so elegant at the surface! :gem:

tests/ckd.rs
```diff
@@ -0,0 +1,98 @@
...
+            *participant,
+            Box::new(keygen::<C>(participants, *participant, threshold, OsRng).unwrap()),
+        ));
+    }
+    run_protocol(protocols).unwrap().into_iter().collect()
+}
+
+#[test]
+fn test_ckd() -> Result<(), Box<dyn Error>> {
+    let mut rng = OsRng;
```
netrome, line 35: Would be nice with a seeded RNG here

tests/ckd.rs
```diff
@@ -0,0 +1,98 @@
...
+
+fn make_keygen(
+    participants: &[Participant],
+    threshold: usize,
+) -> HashMap<Participant, KeygenOutput> {
+    let mut protocols: Vec<ParticipantAndProtocol<KeygenOutput>> = Vec::new();
+    for participant in participants {
+        protocols.push((
+            *participant,
+            Box::new(keygen::<C>(participants, *participant, threshold, OsRng).unwrap()),
```
netrome, line 27: And here

src/test.rs
```diff
@@ -289,3 +289,15 @@ impl RngCore for MockCryptoRng {
...
+pub fn check_common_traits_for_type<T: Clone + Eq + PartialEq + std::fmt::Debug>(v: T) {
+    // Make sure can be debug-printed. This also catches if the Debug does not
+    // have an endless recursion (a popular mistake).
+    println!("{:?}", v);
+    // Test Clone and Eq
+    assert_eq!(v, v.clone());
+    // Make sure it can be unwrapped in a Result (which requires Debug).
+    let e: Result<T, ()> = Ok(v.clone());
+    assert_eq!(v, e.unwrap());
+}
```
netrome, line 303: Interesting check!

src/confidential_key_derivation/protocol.rs
```diff
@@ -64,36 +59,24 @@ async fn do_ckd_coordinator(
...
 
     // Receive everyone's inputs and add them together
     let mut seen = ParticipantCounter::new(&participants);
     let waitpoint = chan.next_waitpoint();
 
     seen.put(me);
     while !seen.full() {
-        let (from, (big_y, big_c)): (_, (CoefficientCommitment, CoefficientCommitment)) =
-            chan.recv(waitpoint).await?;
+        let (from, (big_y, big_c)): (_, (ElementG1, ElementG1)) = chan.recv(waitpoint).await?;
```
netrome, line 74: This is a pretty complex type. Could be worth creating a struct for it.

netrome (changes requested):
Sorry forgot, the doc error is a tiny blocker :sweat_smile:

docs/condidential_key_derivation/confidential_key_derivation.md
```diff
@@ -64,17 +64,19 @@ required.
...
-$\texttt{quote}$, $\texttt{MRTD}$, $\texttt{RTMR0}$, $\texttt{RTMR1}$,
-$\texttt{RTMR2}$, $\texttt{RTMR3}$, *event_log* and *report_data*
-- $`(a,A[=a \cdot G])`$: fresh key pair generated by *app*
+  cryptographic proof that *app* is running inside a genuine TEE. It includes
+  $\texttt{quote}$, $\texttt{MRTD}$, $\texttt{RTMR0}$, $\texttt{RTMR1}$,
+  $\texttt{RTMR2}$, $\texttt{RTMR3}$, *event_log* and *report_data*
+- $`\mathbb{G_1}, \mathbb{G_2}`$ are the groups of the pairing friendly elliptic
+  curves [BLS12_381](https://electriccoin.co/blog/new-snark-curve/), with
+  generators $`G_1, G_2`$ respectively
+- $`(a,A[=a \cdot G_2])`$: fresh key pair in $`\mathbb{G_1}`$ generated by *app*
```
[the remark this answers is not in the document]
gilcu3: good catch!

src/confidential_key_derivation/ciphersuite.rs
```diff
@@ -0,0 +1,309 @@
...
+
+    #[allow(unused)]
+    fn H4(m: &[u8]) -> Self::HashOutput {
+        unimplemented!()
+    }
+
+    #[allow(unused)]
+    fn H5(_m: &[u8]) -> Self::HashOutput {
+        unimplemented!()
+    }
```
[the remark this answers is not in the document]
gilcu3: Yeah, I think the solution should be to eliminate dependence on frost types where we don't really need them, currently we use them in participants, which feels unnecessary.

src/confidential_key_derivation/ciphersuite.rs
```diff
@@ -0,0 +1,309 @@
...
+
+    fn HDKG(
+        m: &[u8],
+    ) -> Option<<<Self::Group as frost_core::Group>::Field as frost_core::Field>::Scalar> {
+        Some(hash_to_scalar(&[CONTEXT_STRING.as_bytes(), b"dkg"], m))
```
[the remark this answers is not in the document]
gilcu3: this strictly follows current frost implementation (inside their crate). So I would rather keep it aligned, if that's ok

src/confidential_key_derivation/ciphersuite.rs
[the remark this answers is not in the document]
netrome: Right yeah. #103 is pretty much spot on, so let's re-open it and do it after we're done with CKD.

src/test.rs
gilcu3: Inspired/taken from frost testing :) maybe I should add the link

tests/ckd.rs
gilcu3: handled elsewhere https://github.com/near/threshold-signatures/pull/101#discussion_r2381903466

src/confidential_key_derivation/ciphersuite.rs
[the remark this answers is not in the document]
netrome: Okay sure :+1:

src/confidential_key_derivation/protocol.rs
netrome: Nice!

SimonRastikian (commented):
Why tests/ckd.rs exist there?
Why does dkg.rs only contains test functions?
Why is ciphersuite.rs in ckd folder?

gilcu3:
> Why tests/ckd.rs exist there? 

That is what is called an integration test. I added it for ckd first, but will also add it for other schemes in a later PR. The idea is that with that you can test the public API of the library, ensuring that users can use the functionality without any other external dependency, and serving as an example on how to use the library.

> Why does dkg.rs only contains test functions? 

This was following the current ecdsa and eddsa implementation, each of which contains the same file. I was planning to open an issue to simplify them, so that there is no need to have so much repeated code testing dkg.

> Why is ciphersuite.rs in ckd folder?

For BLS we need to implement frost_core::Ciphersuite, therefore that file. The main part of this PR is actually there.

docs/condidential_key_derivation/confidential_key_derivation.md
```diff
@@ -84,11 +86,15 @@ $\texttt{RTMR2}$, $\texttt{RTMR3}$, *event_log* and *report_data*
...
+   As a result, $`\texttt{msk}
   = x_1 \cdot λ_1 + \ldots + x_n \cdot λ_n`$, where $λ_i$ are the coefficients
   of the
-  [Lagrange polynomial](https://en.wikipedia.org/wiki/Lagrange_polynomial).
-- $`s`$: the key obtained by *app* as a result of the CKD
+  [Lagrange polynomial](https://en.wikipedia.org/wiki/Lagrange_polynomial). Let
+  $`\texttt{pk} \in \mathbb{G_2}`$ the corresponding public key.
+- $`s`$: the key obtained by *app* as a result of the CKD, which will correspond
+  to a [BLS signature](https://en.wikipedia.org/wiki/BLS_digital_signature) over
+  the hash of $`\texttt{app\_id}`$
```
SimonRastikian, line 97: Please apply an HKDF on s. The secret would be HKDF(s)

docs/condidential_key_derivation/confidential_key_derivation.md
```diff
@@ -162,5 +168,7 @@ contract.
     \texttt{msk} \cdot H(\texttt{app\_id}) + a \cdot Y`$
     - $`\texttt{es} \gets (Y, C) `$
   - Coordinator sends $`\texttt{es}`$ to *app* on-chain
-- *app* obtains $`\texttt{es} = (Y, C)`$ and computes $`s \gets C + (- a) \cdot
-    Y = \texttt{msk} \cdot H(\texttt{app\_id})`$
+- *app* obtains $`\texttt{es} = (Y, C)`$ and computes the BLS signature $`s \gets C + (- a) \cdot  Y`$
+  and checks its correctness with respect to the MPC network public key $`\texttt{pk}`$. If correct,
+  the app can use the computed $`s = \texttt{msk} \cdot H(\texttt{app\_id})`$ as a root deterministic key,
```
SimonRastikian, line 173: I would modify the middle of the sentence to something like  
verifies the signature using the MPC nodes' master public key.

Also please add the HKDF...

SimonRastikian (commented):
I strongly support moving the ciphersuite.rs to the actual one in crypto.
This is a generic ciphersuite and we want to use it potentially in the future for different schemes.

src/confidential_key_derivation/ciphersuite.rs
```diff
@@ -0,0 +1,309 @@
...
+    let element1 = verifying_key.to_element().into();
+    let base2 =
+        <<BLS12381SHA256 as frost_core::Ciphersuite>::Group as frost_core::Group>::generator()
+            .into();
+    let element2 = signature.into();
+    if blstrs::pairing(&base1, &element1) == blstrs::pairing(&element2, &base2) {
+        Ok(())
+    } else {
+        Err(frost_core::Error::InvalidSignature)
+    }
```
SimonRastikian, line 231: Good!

gilcu3:
> I strongly support moving the ciphersuite.rs to the actual one in crypto.
> This is a generic ciphersuite and we want to use it potentially in the future for different schemes.

I agree with that, although this is just the BLS ciphersuite and not general as the one in crypto folder. Looks related to https://github.com/near/threshold-signatures/issues/103, so might be worth doing it when solving that issue about the Ciphersuite trait in general

src/confidential_key_derivation/ciphersuite.rs
```diff
@@ -0,0 +1,309 @@
...
+    G1Projective::hash_to_curve(bytes, DOMAIN, &[])
+}
+
+// From https://github.com/ZcashFoundation/frost/blob/3ffc19d8f473d5bc4e07ed41bc884bdb42d6c29f/frost-secp256k1/src/lib.rs#L161
+fn hash_to_scalar(domain: &[&[u8]], msg: &[u8]) -> blstrs::Scalar {
+    let mut u = [ScalarWrapper(
+        <BLS12381ScalarField as frost_core::Field>::zero(),
+    )];
+    hash_to_field::<ExpandMsgXmd<Sha256>, ScalarWrapper>(&[msg], domain, &mut u)
+        .expect("should never return error according to error cases described in ExpandMsgXmd");
```
SimonRastikian, line 246: Do we want it to fail? can we have an error instead?

Also what do you think about making this a more generic function, maybe add it to the ciphersuite.rs file?

SimonRastikian: Can  we have Sha3 instead? I feel Sha2 is slowly becoming less recommended by the NIST...

src/confidential_key_derivation/ciphersuite.rs
```diff
@@ -0,0 +1,309 @@
...
+
+        let mut d0 = GenericArray::default();
+        d0[8..].copy_from_slice(&data[0..24]);
+        let d0 = blstrs::Scalar::from_bytes_be(&d0.into()).unwrap();
+
+        let mut d1 = GenericArray::default();
+        d1[8..].copy_from_slice(&data[24..]);
+        let d1 = blstrs::Scalar::from_bytes_be(&d1.into()).unwrap();
+
+        ScalarWrapper(d0 * F_2_192 + d1)
```
SimonRastikian, line 273: Could you please comment what this does? I do not understand.

src/confidential_key_derivation/mod.rs
```diff
@@ -9,55 +9,52 @@
...
 }
 
 impl CKDCoordinatorOutput {
-    pub fn new(big_y: Element, big_c: Element) -> Self {
-        CKDCoordinatorOutput {
-            big_y: CoefficientCommitment::new(big_y),
-            big_c: CoefficientCommitment::new(big_c),
-        }
+    pub fn new(big_y: ElementG1, big_c: ElementG1) -> Self {
+        CKDCoordinatorOutput { big_y, big_c }
```
SimonRastikian, line 36: Maybe you need a Self here

src/confidential_key_derivation/mod.rs
```diff
@@ -9,55 +9,52 @@
...
-pub(crate) type Scalar = frost_core::Scalar<Secp256K1Sha256>;
+pub type ElementG1 = blstrs::G1Projective;
+pub type ElementG2 = blstrs::G2Projective;
+pub type Scalar = blstrs::Scalar;
+pub type KeygenOutput = crate::KeygenOutput<BLS12381SHA256>;
+pub type SigningShare = crate::SigningShare<BLS12381SHA256>;
 
 /// The output of the confidential key derivation protocol when run by the coordinator
 #[derive(Debug, Clone, Deserialize, Serialize)]
 pub struct CKDCoordinatorOutput {
```
SimonRastikian, line 29: Call this a Signature now that CKD adds the verifiability.

src/confidential_key_derivation/mod.rs
```diff
@@ -9,55 +9,52 @@
...
-    pub fn unmask(&self, secret_scalar: Scalar) -> CoefficientCommitment {
-        CoefficientCommitment::new(self.big_c.value() - self.big_y.value() * secret_scalar)
+    /// s <- C − a ⋅ Y = msk ⋅ H ( app_id )
+    pub fn unmask(&self, secret_scalar: Scalar) -> Signature {
+        self.big_c - self.big_y * secret_scalar
     }
 }
 
 /// None for participants and Some for coordinator
 pub type CKDOutput = Option<CKDCoordinatorOutput>;
```
SimonRastikian, line 57: Call this SignatureOption

src/confidential_key_derivation/protocol.rs
```diff
@@ -1,57 +1,52 @@
...
     rng: &mut impl CryptoRngCore,
-) -> Result<CKDOutput, ProtocolError> {
+) -> Result<(ElementG1, ElementG1), ProtocolError> {
     // y <- ZZq* , Y <- y * G
-    let (y, big_y) = Secp256K1Sha256::generate_nonce(rng);
+    let y = Scalar::random(rng);
+    let big_y = ElementG1::generator() * y;
     // H(app_id) when H is a random oracle
-    let hash_point = hash2curve(app_id)?;
+    let hash_point = hash2curve(app_id);
```
SimonRastikian, line 24: How come the hash2curve does not fail anymore? Does it panic? if so we need a wrapper not to panic the nodes.

src/confidential_key_derivation/protocol.rs
```diff
@@ -1,57 +1,52 @@
...
-    .map_err(|_| ProtocolError::HashingError)?;
-    Ok(hash)
-}
-
-#[allow(clippy::too_many_arguments)]
-async fn do_ckd_participant(
-    mut chan: SharedChannel,
-    participants: ParticipantList,
-    coordinator: Participant,
+fn ckd_helper(
```
SimonRastikian, line 12: I would align the function name with the other modules function names (e.g., in OT based ECDSA). I would also push the function down in the file (see the other implementations)

tests/ckd.rs
```diff
@@ -0,0 +1,98 @@
...
+    keygen,
+    protocol::{run_protocol, Participant, Protocol},
+};
+
+type C = threshold_signatures::confidential_key_derivation::ciphersuite::BLS12381SHA256;
+type Scalar = BLS12381Scalar;
+
+type ParticipantAndProtocol<T> = (Participant, Box<dyn Protocol<Output = T>>);
+
+fn make_keygen(
```
SimonRastikian, line 19: Maybe call this run_keygen?

tests/ckd.rs
```diff
@@ -0,0 +1,98 @@
...
+    let app_sk = Scalar::random(&mut rng);
+    let app_pk = G1Projective::generator() * app_sk;
+
+    // create participants
+    let threshold = 3;
+    let participants = vec![
+        Participant::from(0u32),
+        Participant::from(1u32),
+        Participant::from(2u32),
+    ];
```
SimonRastikian, line 48: There is a function that can do this, please use it.

tests/ckd.rs
```diff
@@ -0,0 +1,98 @@
...
+
+    let result = make_keygen(&participants, threshold);
+
+    assert!(result.len() == participants.len());
+
+    let public_key = result.get(&participants[0]).unwrap().public_key;
```
SimonRastikian, line 54: I doubt you need all this. you can probably use result[0].public_key.

tests/ckd.rs
```diff
@@ -0,0 +1,98 @@
...
+
+    // choose a coordinator at random
+    let index = OsRng.next_u32() % participants.len() as u32;
+    let coordinator = participants[index as usize];
```
SimonRastikian, line 58: Maybe it's nice to create a generic function in src/test.rs that does that as the code is now redundant in several places.

src/confidential_key_derivation/mod.rs
gilcu3: What do you mean? If it compiles then certainly it is valid as written. Anyway I think this was fixed already, as pedantic clippy preferred using Self :)

src/confidential_key_derivation/ciphersuite.rs
gilcu3: > Do we want it to fail? can we have an error instead?

It should never fail, that's why the `expect`. This is implemented exactly like that in other libraries as well, following the standard which does not expect any failure.

> Can we have Sha3 instead? 

Using sha2 is consistent with the current standard https://www.rfc-editor.org/rfc/rfc9380.html#name-bls12-381-g1-2, so I think keeping like this for now should be preferable.

src/confidential_key_derivation/mod.rs
gilcu3: Let's leave this as a follow-up discussion as I mentioned https://github.com/near/threshold-signatures/pull/101#discussion_r2394130820

tests/ckd.rs
gilcu3: Will handle in https://github.com/near/threshold-signatures/pull/127

src/confidential_key_derivation/protocol.rs
gilcu3: Alright, found correct function name :) `compute_signature_share`

Issue #18, opened 2025-07-23
Implement a "hardcoded" restriction over the threshold in the different schemes


SimonRastikian:
### Background

The threshold may differ in some schemes whether DKG, robust ECDSA or OT based ECDSA. The threshold should not exceed in some cases the 1/3 of participants and in other cases can be almost every value less than N.
Solution Hardcode some restriction in this threshold signatures repo and potentially add those restrictions to the MPC repo (redundancy in this case does not hurt)

SimonRastikian:
Sorry for taking some time, I was very busy lately.
I think the idea is different. It's a bit complex but I try to explain it plainly.

First for DKG:  voting requires, due to broadcast,  that the `max_faulty_nodes` should be 33% of the total number of nodes.
                       the secret shares generated however would have the threshold = `max_malicious_parties` + 1

In Ot-based ecdsa: generating a signature would require at least  `max_malicious_parties` + 1 which this would be named as threshold

In  robust-ecdsa: we will use the same terminology as the original paper meaning that `threshold = max_malicious_parties` but where one needs at least 2*`max_malicious_parties` + 1 nodes to be active to generate a valid signature.

Not sure fully how this would be designed as interface but things must be done there

PR #106, opened 2025-09-25
feat(threshold): add module to validate cryptographic parameters

0xsecaas:
A a module for validating cryptographic scheme parameters (`N`, `f`) and deriving the `threshold`.

It provides a single `validate_and_derive_threshold` function to ensure protocol safety for `DKG`, `OT-based ECDSA`, `Robust ECDSA` and possible future schemes.

related to #18


0xsecaas:
Something is off, please let me review first.

0xsecaas (commented):
Ok, ready to review.

src/ecdsa/ot_based_ecdsa/test.rs
```diff
@@ -166,10 +167,10 @@ fn test_refresh() -> Result<(), Box<dyn Error>> {
     let participants = generate_participants(11);
     let max_malicious = 5;
     let threshold = max_malicious + 1;
-    let keys = run_keygen(&participants, threshold)?;
+    let keys = run_keygen(Scheme::OtBasedEcdsa, &participants, threshold)?;
```
SimonRastikian: The "t" is capital OT refers to Oblivious Transfer.
```suggestion
    let keys = run_keygen(Scheme::OTBasedEcdsa, &participants, threshold)?;
```

src/threshold.rs
```diff
@@ -0,0 +1,363 @@
+//! Threshold-cryptography parameter validation.
+//!
+//! Provides a safe API to validate the number of faulty nodes (`f`) for a given number of
+//! Participants (`N`) and derive the canonical threshold (`t`) for different cryptography schemes.
+//!
+//! # Parameters
+//!
+//! - **`N`**: Total participants.
+//! - **`f`**: Maximum assumed faulty participants.
```
SimonRastikian: In distributed systems, the terminology used is "faulty" nodes. In cryptography, the terminology is malicious nodes.

src/threshold.rs
```diff
@@ -0,0 +1,363 @@
...
+//! - **`t` (threshold)**: Minimum participants required to generate a signature, derived from `N` and `f`.
```
SimonRastikian: Let us not use the term threshold as multiple definitions exist in the wild and we do not want to confuse the library users.

src/threshold.rs
```diff
@@ -0,0 +1,363 @@
...
+    }
+
+    let threshold = match scheme {
+        Scheme::Dkg => {
+            // require f <= floor(n/3) which is equivalent to 3f <= n
+            if f.checked_mul(3).ok_or(ValidationError::ArithmeticError)? > n {
+                return Err(ValidationError::FTooLargeForDkg);
+            }
+            f.checked_add(1).ok_or(ValidationError::ArithmeticError)?
+        }
```
SimonRastikian: DKG is very tricky as there are two actual "thresholds". The first is the Distributed systems' threshold which assumes that not more than one third of the participants are malicious (while running the DKG). The second is the threshold you are creating for the cryptography part i.e., the minimal number of shares needed to recover the key. Both exist in the DKG. 
One might ask whether it makes sense to have these two thresholds different. The only answer I can give is "the library user should be clearly notified about this thing and should know what he/she is doing"

src/threshold.rs
```diff
@@ -0,0 +1,363 @@
...
+
+/// Supported cryptographic schemes.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum Scheme {
+    /// Distributed Key Generation
+    Dkg,
+    /// OT-based ECDSA
+    OtBasedEcdsa,
+    /// Robust ECDSA
+    RobustEcdsa,
```
SimonRastikian: What about parameters for the subprotocols (presign, sign and so on?) Some of the schemes should be robust and "might" allow changing some parameters from the one protocol to the other.

PR #111, opened 2025-09-29
chore: refactor dkg tests

### LINKED ISSUE #110: Redundance in DKG tests
### Background

Currently we have the same tests for DKG in ecdsa, eddsa and ckd. Given that the only difference is generic, they could be simplified such that they are only implemented once.

### User Story

As a user I would prefer to have less code redundancy

### Acceptance Criteria

Code is reused between DKG tests.

_No response_

gilcu3:
Closes #110


netrome (approved):
Nice stuff! One unrelated comment.

src/confidential_key_derivation/test.rs
```diff
@@ -0,0 +1,18 @@
...
+
+#[test]
+fn test_refresh() -> Result<(), Box<dyn Error>> {
+    crate::dkg::test::test_refresh::<C>()
+}
+
+#[test]
+fn test_reshare() -> Result<(), Box<dyn Error>> {
+    crate::dkg::test::test_reshare::<C>()
+}
```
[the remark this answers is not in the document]
gilcu3: Yeah, we need to do that globally for sure. I am actually liking rust tests a lot, free use of unwrap and expect :dagger:

src/dkg.rs
```diff
@@ -661,4 +667,113 @@ mod test {
...
+    pub fn test_keygen<C: Ciphersuite>() -> Result<(), Box<dyn Error>>
+    where <C::Group as frost_core::Group>::Element: std::fmt::Debug + std::marker::Send,
+    <<<C as frost_core::Ciphersuite>::Group as frost_core::Group>::Field as frost_core::Field>::Scalar: std::marker::Send
+    {
+        let participants = vec![
+            Participant::from(31u32),
+            Participant::from(1u32),
+            Participant::from(2u32),
+        ];
+        let threshold = 3;
```
SimonRastikian, line 680: Why is the threshold the same as the number of participants? Does this mean there are 2 malicious actors out of 3?
I would make both numbers as input.

src/dkg.rs
```diff
@@ -661,4 +667,113 @@ mod test {
...
+        let participants = vec![result[0].0, result[1].0, result[2].0];
+        let shares = [
+            result[0].1.private_share.to_scalar(),
+            result[1].1.private_share.to_scalar(),
+            result[2].1.private_share.to_scalar(),
+        ];
+        let p_list = ParticipantList::new(&participants).unwrap();
+        let x = p_list.lagrange::<C>(participants[0])? * shares[0]
+            + p_list.lagrange::<C>(participants[1])? * shares[1]
+            + p_list.lagrange::<C>(participants[2])? * shares[2];
```
SimonRastikian, line 697: Please adapt this according to the input number of participants and threshold (using a loop)

src/dkg.rs
```diff
@@ -661,4 +667,113 @@ mod test {
...
+    pub fn test_refresh<C: Ciphersuite>() -> Result<(), Box<dyn Error>>
+    where <C::Group as frost_core::Group>::Element: std::fmt::Debug + std::marker::Send,
+    <<<C as frost_core::Ciphersuite>::Group as frost_core::Group>::Field as frost_core::Field>::Scalar: std::marker::Send
+    {
+        let participants = vec![
+            Participant::from(0u32),
+            Participant::from(31u32),
+            Participant::from(2u32),
+        ];
+        let threshold = 3;
```
SimonRastikian, line 711: Same as before

src/dkg.rs
```diff
@@ -661,4 +667,113 @@ mod test {
...
+
+        let result0 = run_keygen::<C>(&participants, threshold)?;
+        assert_public_key_invariant(&result0);
+
+        let pub_key = result0[2].1.public_key.to_element();
```
SimonRastikian, line 716: Why index 2? any specific reason?

src/dkg.rs
```diff
@@ -661,4 +667,113 @@ mod test {
...
+    }
+
+    pub fn test_reshare<C: Ciphersuite>() -> Result<(), Box<dyn Error>>
+    where <C::Group as frost_core::Group>::Element: std::fmt::Debug + std::marker::Send,
+    <<<C as frost_core::Ciphersuite>::Group as frost_core::Group>::Field as frost_core::Field>::Scalar: std::marker::Send
+    {
+        let participants = generate_participants(3);
+        let threshold0 = 2;
+        let threshold1 = 3;
+
```
SimonRastikian, line 742: Same. Please as inputs.

src/dkg.rs
```diff
@@ -661,4 +667,113 @@ mod test {
...
+            result1[0].1.private_share.to_scalar(),
+            result1[1].1.private_share.to_scalar(),
+            result1[2].1.private_share.to_scalar(),
+            result1[3].1.private_share.to_scalar(),
+        ];
+        let p_list = ParticipantList::new(&participants).unwrap();
+        let x = p_list.lagrange::<C>(participants[0])? * shares[0]
+            + p_list.lagrange::<C>(participants[1])? * shares[1]
+            + p_list.lagrange::<C>(participants[2])? * shares[2]
+            + p_list.lagrange::<C>(participants[3])? * shares[3];
```
SimonRastikian, line 771: Can we have a for loop and a code relaxation?

x += p_list.lagrange::<C>(participants[i])? * result1[i].1.private_share.to_scalar(),

src/confidential_key_derivation/test.rs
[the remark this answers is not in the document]
gilcu3: > However, propagating an error tells you more about that.

I thought it was the opposite. That expect and unwrap will tell you the exact line of failure, which propagating will lose that.

src/dkg.rs
gilcu3: None really, it was like that before. Will change to 0 if that makes more sense

netrome (approved):
Still looks good to me.

netrome (approved):

src/dkg.rs
```diff
@@ -668,82 +667,62 @@ pub mod test {
...
-        let threshold = 3;
-
-        let result = run_keygen::<C>(&participants, threshold)?;
+        let result = run_keygen::<C>(&participants, threshold).unwrap();
         assert!(result.len() == participants.len());
         assert_public_key_invariant(&result);
 
-        let pub_key = result[2].1.public_key.to_element();
+        let pub_key = result[0].1.public_key.to_element();
+        let participants = result.iter().map(|p| p.0).collect::<Vec<_>>();
```
SimonRastikian, line 677: Why do you need to collect again? you have those participants as input no?

SimonRastikian: If it's for the order just do participants.sort()

src/dkg.rs
```diff
@@ -668,82 +667,62 @@ pub mod test {
...
+        let participants = result1.iter().map(|p| p.0).collect::<Vec<_>>();
+        let shares = result1
+            .iter()
+            .map(|r| r.1.private_share.to_scalar())
+            .collect::<Vec<_>>();
         let p_list = ParticipantList::new(&participants).unwrap();
-        let x = p_list.lagrange::<C>(participants[0])? * shares[0]
-            + p_list.lagrange::<C>(participants[1])? * shares[1]
-            + p_list.lagrange::<C>(participants[2])? * shares[2];
+        let mut x = <<<C as frost_core::Ciphersuite>::Group as frost_core::Group>::Field as frost_core::Field>::zero();
```
SimonRastikian, line 709: I think you can do `<C::Group as Group>::Field::zero()` instead

src/dkg.rs
gilcu3: Unfortunately cannot be done that way, because this requires muting participants, which is behind a reference.

src/dkg.rs
gilcu3: Something similar yes :) , fixed in 1c192b90fc7798cf36270df1cea84aab6e590025

PR #113, opened 2025-09-29
Simon/test robustness

SimonRastikian:
This PR helps testing that our schemes are indeed robust, meaning if one generates presignatures with more participants than actually needed, then fewer presignatures elements can be used during signing and everything would succeed.

### DIFF
--- a/src/ecdsa/ot_based_ecdsa/test.rs
+++ b/src/ecdsa/ot_based_ecdsa/test.rs
@@ -1,7 +1,7 @@
 use super::{
     presign::presign,
     sign::sign,
-    triples::{test::deal, TriplePub, TripleShare},
+    triples::{generate_triple_many, test::deal, TriplePub, TripleShare},
     PresignArguments, PresignOutput,
 };
 use crate::ecdsa::{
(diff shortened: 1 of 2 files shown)


src/ecdsa/ot_based_ecdsa/test.rs
```diff
@@ -327,3 +327,132 @@ fn test_e2e_random_identifiers_with_rerandomization() -> Result<(), Box<dyn Erro
...
+    let mut odd = Vec::with_capacity(v.len() / 2);
+    for (i, x) in v.iter().enumerate() {
+        if i % 2 == 0 {
+            even.push(x.clone());
+        } else {
+            odd.push(x.clone());
+        }
+    }
+    (even, odd)
+}
```
netrome, line 342: Nit: While this is valid - you can also use the `partition` method on iterators for this:

```suggestion
fn split_even_odd<T: Clone>(v: Vec<T>) -> (Vec<T>, Vec<T>) {
    v.into_iter().enumerate().partition(|(i, _)| i % 2 == 0).map(|(_, x)| x)
}
```

src/ecdsa/ot_based_ecdsa/test.rs
```diff
@@ -327,3 +327,132 @@ fn test_e2e_random_identifiers_with_rerandomization() -> Result<(), Box<dyn Erro
...
 }
+
+fn split_even_odd<T: Clone>(v: Vec<T>) -> (Vec<T>, Vec<T>) {
+    let mut even = Vec::with_capacity(v.len() / 2 + 1);
+    let mut odd = Vec::with_capacity(v.len() / 2);
+    for (i, x) in v.iter().enumerate() {
+        if i % 2 == 0 {
+            even.push(x.clone());
+        } else {
+            odd.push(x.clone());
```
netrome, line 338: If you don't go with the `partition` suggestion, you can skip the clones by using `into_iter` here

```suggestion
    for (i, x) in v.into_iter().enumerate() {
        if i % 2 == 0 {
            even.push(x);
        } else {
            odd.push(x);
```

src/ecdsa/ot_based_ecdsa/test.rs
```diff
@@ -327,3 +327,132 @@ fn test_e2e_random_identifiers_with_rerandomization() -> Result<(), Box<dyn Erro
...
+
+    let mut two_triples = run_protocol(protocols)?;
+    two_triples.sort_by_key(|(p, _)| *p);
+
+    let (shares, pubs): (Vec<_>, Vec<_>) = two_triples.into_iter().flat_map(|(_, vec)| vec).unzip();
+
+    // split shares into shares0 and shares 1 and pubs into pubs0 and pubs1
+    let (mut shares0, mut shares1) = split_even_odd(shares);
+
+    // split shares into shares0 and shares 1 and pubs into pubs0 and pubs1
```
netrome, line 377: I don't find these comments helpful
```suggestion
    let (mut shares0, mut shares1) = split_even_odd(shares);
```

src/ecdsa/ot_based_ecdsa/test.rs
```diff
@@ -327,3 +327,132 @@ fn test_e2e_random_identifiers_with_rerandomization() -> Result<(), Box<dyn Erro
...
+        &pub1[0],
+        threshold,
+    )?;
+    let msg = b"hello world";
+
+    // Use less presignatures to sign
+    presign_result.remove(0);
+
+    run_sign_without_rerandomization(presign_result, public_key, msg)?;
+    Ok(())
```
netrome, line 399: Okay so we're both testing robustness for triplets -> presignatures. I.e. we remove the triplets of one participant before making the presignature, and we're removing one presignature before running signing?

src/ecdsa/ot_based_ecdsa/test.rs
```diff
@@ -327,3 +327,132 @@ fn test_e2e_random_identifiers_with_rerandomization() -> Result<(), Box<dyn Erro
...
+        &pub0[0],
+        &pub1[0],
+        threshold,
+    )?;
+    let msg = b"hello world";
+
+    // Use less presignatures to sign
+    presign_result.remove(0);
+
+    run_sign_without_rerandomization(presign_result, public_key, msg)?;
```
netrome, line 398: It would be nice to add a validation step here to see that the returned signature is valid.

src/ecdsa/ot_based_ecdsa/test.rs
```diff
@@ -327,3 +327,132 @@ fn test_e2e_random_identifiers_with_rerandomization() -> Result<(), Box<dyn Erro
...
+        threshold,
+    )?;
+    let msg = b"hello world";
+
+    // Use less presignatures to sign
+    presign_result.remove(0);
+
+    run_sign_with_rerandomization(presign_result, public_key, msg)?;
+    Ok(())
+}
```
netrome, line 458: From what I can see, both of these tests are identical except the sign function used. While some code duplication in tests isn't a big issue imo - we could consider factoring out the shared logic like

```suggestion
#[test]
fn test_robustness_without_rerandomization() {
    test_robustness(run_sign_without_rerandomization).expect("robustness test should succeed");
}

#[test]
fn test_robustness_with_rerandomization() {
    test_robustness(run_sign_with_rerandomization).expect("robustness test should succeed");
}

fn test_robustness<T>(sign_fn: SignFunction<T>) -> Result<T, Box<dyn Error>> {
    let participants_count = 7;
    let mut rng = OsRng;
    let mut participants = generate_participants_with_random_ids(participants_count, &mut rng);
    let threshold = 4;
    let mut key_packages = run_keygen::<Secp256K1Sha256>(&participants.clone(), threshold)?;
    assert_public_key_invariant(&key_packages);

    let public_key = key_packages[0].1.public_key.to_element();
    // Now remove a participant
    // You can remove the same index because both participants and key_packages are sorted in the same way
    participants.remove(0);
    key_packages.remove(0);

    let mut protocols: Vec<(_, Box<dyn Protocol<Output = _>>)> =
        Vec::with_capacity(participants.len());

    // Generate triples with 6 participants
    for &p in &participants {
        let protocol = generate_triple_many::<2>(&participants, p, threshold, rng);
        let protocol = protocol.unwrap();
        protocols.push((p, Box::new(protocol)));
    }

    let mut two_triples = run_protocol(protocols)?;
    two_triples.sort_by_key(|(p, _)| *p);

    let (shares, pubs): (Vec<_>, Vec<_>) = two_triples.into_iter().flat_map(|(_, vec)| vec).unzip();

    // split shares into shares0 and shares 1 and pubs into pubs0 and pubs1
    let (mut shares0, mut shares1) = split_even_odd(shares);

    // split shares into shares0 and shares 1 and pubs into pubs0 and pubs1
    let (pub0, pub1) = split_even_odd(pubs);

    // Test robustness for presig with less triples than originally generated
    key_packages.remove(0);
    shares0.remove(0);
    shares1.remove(0);

    let mut presign_result = run_presign(
        key_packages,
        shares0,
        shares1,
        &pub0[0],
        &pub1[0],
        threshold,
    )?;
    let msg = b"hello world";

    // Use less presignatures to sign
    presign_result.remove(0);

    sign_fn(presign_result, public_key, msg)
}

type SignFunction<'a, T> = fn(Participants, PubKey, Msg<'a>) -> Result<T, Box<dyn Error>>;
type Participants = Vec<(Participant, PresignOutput)>;
type PubKey = Element;
type Msg<'a> = &'a [u8];
```

Here I also changed the top level test functions to not return results. It would be nice to do the same for the `test_robustness` one as well.

netrome (approved):
Nice stuff. Only code suggestions from my side. Great to have tests for this.

src/ecdsa/ot_based_ecdsa/test.rs
SimonRastikian: Done!

src/ecdsa/ot_based_ecdsa/test.rs
gilcu3, line 398: nit: a small comment about what should fail if there was no robustness would make the test clear

gilcu3 (approved):
Nice!

Left a couple of comments/questions.

src/ecdsa/ot_based_ecdsa/test.rs
netrome: Right yeah my suggestion is wrong. Sure though this suggestion while a bit cleaner is less efficient since it requires two passes of the same data.

src/ecdsa/ot_based_ecdsa/test.rs
gilcu3: Suggestions by AI with one pass :) (not tested)

```rust
fn split_even_odd<T: Clone>(v: Vec<T>) -> (Vec<T>, Vec<T>) {
    v.into_iter()
        .enumerate()
        .partition::<Vec<_>, _>(|(i, _)| i % 2 == 0)
        .into_iter()
        .map(|(evens, odds)| {
            (
                evens.into_iter().map(|(_, x)| x).collect(),
                odds.into_iter().map(|(_, x)| x).collect(),
            )
        })
        .next()
        .unwrap_or_else(|| (Vec::new(), Vec::new()))
}
```

Or, more simply, using `fold`:

```rust
fn split_even_odd<T: Clone>(v: Vec<T>) -> (Vec<T>, Vec<T>) {
    v.into_iter().enumerate().fold((Vec::new(), Vec::new()), |(mut even, mut odd), (i, x)| {
        if i % 2 == 0 {
            even.push(x);
        } else {
            odd.push(x);
        }
        (even, odd)
    })
}
```

src/ecdsa/ot_based_ecdsa/test.rs
SimonRastikian: Haha I do not think any of the two options is actually better in terms of complexity.
I will keep it as is.

src/ecdsa/ot_based_ecdsa/test.rs
SimonRastikian: The validation happens internally to the run.

src/ecdsa/ot_based_ecdsa/test.rs
```diff
@@ -327,3 +327,132 @@ fn test_e2e_random_identifiers_with_rerandomization() -> Result<(), Box<dyn Erro
...
+            even.push(x.clone());
+        } else {
+            odd.push(x.clone());
+        }
+    }
+    (even, odd)
+}
+
+#[test]
+fn test_robustness_without_rerandomization() -> Result<(), Box<dyn Error>> {
```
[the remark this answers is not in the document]
SimonRastikian: I would have thought the same as the error could be many different things.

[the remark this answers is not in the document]
SimonRastikian: Done!

netrome (approved):
Nice stuff!

PR #115, opened 2025-09-29
feat: add hash_to_field following bls12_381 crate

### LINKED ISSUE #105: Implement proper hash_to_field function
### Background

This is a follow up to https://github.com/near/threshold-signatures/pull/101
There, a toy hash_to_field function was used.

### User Story

Use well proved and standardized functions.

### Acceptance Criteria

The implemented function is taken from a well known standard or implementation. If test vectors exist, these should be used in tests as well.

_No response_

gilcu3:
Closes #105

### DIFF
--- a/src/confidential_key_derivation/ciphersuite.rs
+++ b/src/confidential_key_derivation/ciphersuite.rs
@@ -1,3 +1,5 @@
+use std::sync::LazyLock;
+
 use digest::{consts::U48, generic_array::GenericArray};
 use elliptic_curve::hash2curve::{hash_to_field, ExpandMsgXmd, FromOkm};
 use rand_core::{CryptoRng, RngCore};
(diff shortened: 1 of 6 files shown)


src/confidential_key_derivation/ciphersuite.rs
```diff
@@ -247,45 +247,61 @@ fn hash_to_scalar(domain: &[&[u8]], msg: &[u8]) -> blstrs::Scalar {
...
+impl ScalarWrapper {
+    // Based on https://github.com/arkworks-rs/algebra/blob/c6f9284c17df00c50d954a5fe1c72dd4a5698103/ff/src/fields/prime.rs#L72
+    // Converts `bytes` into a `Scalar`
+    fn from_le_bytes_mod_order(bytes: &[u8]) -> Self {
+        let mut res = blstrs::Scalar::ZERO;
+
+        let mut count = 0;
+        let mut remainder = 0;
+        for byte in bytes.iter().rev() {
+            remainder = (remainder << 8) + u64::from(*byte);
```
netrome, line 262: This feels like it could overflow. Please use `checked_add` instead, or explicitly use `overflowing_add` if that's the intended behavior.

src/confidential_key_derivation/ciphersuite.rs
```diff
@@ -247,45 +247,61 @@ fn hash_to_scalar(domain: &[&[u8]], msg: &[u8]) -> blstrs::Scalar {
...
+            count += 1;
+            if count == 8 {
+                res = res.shl(64) + blstrs::Scalar::from(remainder);
```
netrome, line 265: Same here

netrome (changes requested):
I would like to get rid of potentially overflowing arithmetic before merging this. Also, it could be nice to try to trigger the overflow in a test case if it's easy to do.

src/confidential_key_derivation/ciphersuite.rs
```diff
@@ -247,45 +247,61 @@ fn hash_to_scalar(domain: &[&[u8]], msg: &[u8]) -> blstrs::Scalar {
...
-        ])
-        .unwrap();
-
-        let mut d0 = GenericArray::default();
-        d0[8..].copy_from_slice(&data[0..24]);
-        let d0 = blstrs::Scalar::from_bytes_be(&d0.into()).unwrap();
+impl ScalarWrapper {
+    // Based on https://github.com/arkworks-rs/algebra/blob/c6f9284c17df00c50d954a5fe1c72dd4a5698103/ff/src/fields/prime.rs#L72
+    // Converts `bytes` into a `Scalar`
+    fn from_le_bytes_mod_order(bytes: &[u8]) -> Self {
```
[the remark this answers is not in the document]
gilcu3: I can do it here. Do you propose using a fuzzer or just a manual test trying random values?

src/confidential_key_derivation/ciphersuite.rs
netrome: Nvm these are u8s so this should never overflow.

src/confidential_key_derivation/ciphersuite.rs
gilcu3: This is modular arithmetic be `blstrs::Scalar`, so it cannot overflow either.

src/confidential_key_derivation/ciphersuite.rs
gilcu3: does not overflow https://nearone.slack.com/archives/C0912BTG51T/p1759303992795069?thread_ts=1759169997.806039&cid=C0912BTG51T

src/confidential_key_derivation/ciphersuite.rs
netrome: Right I missed that.

src/confidential_key_derivation/ciphersuite.rs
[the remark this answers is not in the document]
gilcu3: For my own peace of mind, I will go here with the poor's man fuzzer, we can do it properly in a follow up as you suggested

src/confidential_key_derivation/ciphersuite.rs
```diff
@@ -247,45 +247,63 @@ fn hash_to_scalar(domain: &[&[u8]], msg: &[u8]) -> blstrs::Scalar {
...
+impl FromOkm for ScalarWrapper {
+    // ceil(log2(p)) = 255, m = 1, k = 128.
+    type Length = U48;
 
-        ScalarWrapper(d0 * F_2_192 + d1)
+    fn from_okm(okm: &GenericArray<u8, Self::Length>) -> Self {
+        let mut bs = [0u8; 64];
+        bs[16..].copy_from_slice(okm);
+        bs.reverse(); // into little endian
+        ScalarWrapper::from_le_bytes_mod_order(&bs)
```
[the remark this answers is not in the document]
gilcu3: For this I don't have much comments except what is already mentioned there, with the link to the source of the code upstream. Looking at it again, we might not need to do that, will make a quick check and get back here.

src/confidential_key_derivation/ciphersuite.rs
```diff
@@ -296,14 +314,67 @@ mod tests {
...
+    // This is guaranteed by the `test_verify_overflow_failure` below
+    fn test_stress_test_scalarwrapper_from_le_bytes_mod_order() {
+        // empty case
+        ScalarWrapper::from_le_bytes_mod_order(&[]);
+        let mut rng = rand::rngs::OsRng;
+        for _ in 0..1000 {
+            let len = rng.gen_range(1..10000);
+            let mut bytes = vec![0; len];
+            rng.fill_bytes(&mut bytes);
+            ScalarWrapper::from_le_bytes_mod_order(&bytes);
```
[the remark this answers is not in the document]
gilcu3: Not really. I though of an indirect way, using values from the standard, but as they usually have 2 scalars as output instead of one (in the test vectors), just testing that would require having a bit more redundant code here. If you think it is worth it, I am more than glad to implement it :)

src/confidential_key_derivation/ciphersuite.rs
```diff
@@ -247,45 +247,63 @@ fn hash_to_scalar(domain: &[&[u8]], msg: &[u8]) -> blstrs::Scalar {
...
+                remainder = 0;
+                count = 0;
+            }
+        }
+        if count > 0 {
+            res = res * res.shl(count * 8) + blstrs::Scalar::from(remainder);
+        }
+        Self(res)
+    }
+}
```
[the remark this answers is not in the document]
gilcu3: done in c8e25eca0e124b2daff075896bc56011fe560ec0

PR #117, opened 2025-09-29
Simon/no panic participants index

SimonRastikian:
This PR prevents panic when searching for an index in ParticipantList and in ParticipantMap.

When I was trying to write unit test case by TrailOfBits in echo_broadcast.rs,
I couldn't help but improve the code readibility by reducing code redundancy in test cases

### DIFF
--- a/src/participants.rs
+++ b/src/participants.rs
@@ -280,3 +280,27 @@ impl<'a> ParticipantCounter<'a> {
         self.counter == 0
     }
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::test::generate_participants;
+
+    #[test]
+    fn test_get_index_participant_error() {
+        let participants = generate_participants(5);
+        let participants = ParticipantList::new(&participants).unwrap();
+        assert!(participants.index(&Participant::from(1234_u32)).is_err())
+    }
+
+    #[test]
+    fn test_get_index_data_error() {
+        let participants = generate_participants(5);
+        let participants = ParticipantList::new(&participants).unwrap();
+        let map: ParticipantMap<'_, u32> = ParticipantMap::new(&participants);
+        // no participant test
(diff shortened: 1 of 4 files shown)


SimonRastikian:
The test cases in echo_broadcast.rs is a code refactor. One can split it into a mini-PR but I do not think it is a big deal as one would have to merge the PRs one after another.

src/participants.rs
```diff
@@ -69,8 +69,13 @@ impl ParticipantList {
...
     /// Basically, the order they appear in a sorted list
-    pub fn index(&self, participant: &Participant) -> usize {
-        self.indices[participant]
+    pub fn index(&self, participant: &Participant) -> Result<usize, ProtocolError> {
+        self.indices
+            .get(participant)
+            .copied()
+            .ok_or(ProtocolError::Other(
+                "Could not find the searched participant".to_string(),
+            ))
```
gilcu3, line 78: nit: We have this error `InvalidParticipantIndex` defined (and unused). I think it fits here.

src/protocol/echo_broadcast.rs
```diff
@@ -432,32 +430,15 @@ mod test {
         Ok(vote_list)
     }
 
-    #[allow(clippy::type_complexity)]
-    pub fn do_broadcast_dishonest_v1(
+    fn do_broadcast_dishonest<F, Fut>(
```
gilcu3, line 433: Nice function unification :)

src/participants.rs
SimonRastikian: Absolutely!

SimonRastikian: I don't like it's in InitializationError

netrome: Should we add the same variant to `ProtocolError` then?

SimonRastikian: Added InvalidIndex

PR #129, opened 2025-10-01
Reducing code redundancy when receiving elements #69

### LINKED ISSUE #69: Reducing code redundancy when receiving elements
Currently there are a lot of repetition in the code in several places when receiving elements e.g., receiving scalars.
It would make sense to have different function helpers that would do the job instead.
Reference see: https://github.com/near/threshold-signatures/pull/63#discussion_r2333547239

0xsecaas:
closes #69

### DIFF
--- a/src/participants.rs
+++ b/src/participants.rs
@@ -201,7 +201,7 @@ impl From<ParticipantList> for Vec<Participant> {
 ///
 /// The idea is that you have one element for each participant.
 #[derive(Debug, Clone, Serialize)]
-pub struct ParticipantMap<'a, T> {
+pub(crate) struct ParticipantMap<'a, T> {
     #[serde(skip_serializing)]
     participants: &'a ParticipantList,
     data: Vec<Option<T>>,
--- a/src/protocol/mod.rs
+++ b/src/protocol/mod.rs
@@ -5,7 +5,7 @@
 //! to deliver messages to and from that protocol, and eventually it will produce
 //! a result, without you having to worry about how many rounds it has, or how
 //! to serialize the emssages it produces.
-pub mod echo_broadcast;
+pub(crate) mod echo_broadcast;
 pub(crate) mod helpers;
 pub(crate) mod internal;
 
--- a/src/protocol/echo_broadcast.rs
+++ b/src/protocol/echo_broadcast.rs
@@ -103,7 +103,7 @@ where
 /// of the broadcasts be the same as the input vote.
 /// `Reliable_broadcast_receive_all` is expected to be called right after `reliable_broadcast_send`.
 #[allow(clippy::too_many_lines)]
-pub(crate) async fn reliable_broadcast_receive_all<'a, T>(
+pub async fn reliable_broadcast_receive_all<'a, T>(
     chan: &SharedChannel,
     wait: Waitpoint,
     participants: &'a ParticipantList,


src/generic_dkg.rs
```diff
@@ -476,15 +486,14 @@ async fn do_keyshare<C: Ciphersuite>(
...
-            chan.recv(wait_round_3).await?;
-        if !seen.put(from) {
-            continue;
-        }
-
+    for (from, signing_share_from) in recv_from_many(
+        &mut chan,
+        wait_round_3,
+        participants.participants(),
+        Some(&[me]),
```
SimonRastikian, line 493: Since almost everywhere we are using this as input do you think it makes sense to create a function
recv_from_others that takes me instead of a list and just receives internally from participants.other(me)?

src/generic_dkg.rs
```diff
@@ -476,15 +486,14 @@ async fn do_keyshare<C: Ciphersuite>(
...
-    let mut seen = ParticipantCounter::new(&participants);
-    seen.put(me);
-    while !seen.full() {
-        let (from, signing_share_from): (Participant, SigningShare<C>) =
-            chan.recv(wait_round_3).await?;
-        if !seen.put(from) {
-            continue;
-        }
-
+    for (from, signing_share_from) in recv_from_many(
```
SimonRastikian, line 489: I am impressed that one does not need to explicitly specify the type of  `signing_share_from`. Any thoughts about it?

src/protocol/helpers.rs
```diff
@@ -0,0 +1,48 @@
...
+            _ => {}
+        }
+    }
+}
+
+/// Gather exactly one message from each participant in a group before proceeding.
+pub async fn recv_from_many<T>(
+    chan: &mut SharedChannel,
+    wait: u64,
+    participants: &[Participant],
```
SimonRastikian, line 25: Maybe let's better tailor the function to the use cases. Instead of having &[Participant] we input ParticipantList?

src/ecdsa/ot_based_ecdsa/triples/generation.rs
```diff
@@ -730,34 +717,28 @@ async fn do_generation_many<const N: usize>(
...
+            Vec<PolynomialCommitment>,
+            Vec<PolynomialCommitment>,
+            Vec<PolynomialCommitment>,
+            Vec<Randomness>,
+            Vec<dlog::Proof<C>>,
+            Vec<dlog::Proof<C>>,
+        )>(&mut chan, wait2, participants.participants(), Some(&[me]))
+        .await?
+        {
+            //#[allow(clippy::type_complexity)]
```
SimonRastikian, line 741: Can you please delete this line?

src/ecdsa/ot_based_ecdsa/triples/generation.rs
```diff
@@ -823,17 +804,15 @@ async fn do_generation_many<const N: usize>(
...
-            }
+        for (_, (a_j_i_v, b_j_i_v)) in recv_from_many::<(
+            Vec<SerializableScalar<C>>,
+            Vec<SerializableScalar<C>>,
+        )>(
+            &mut chan, wait3, participants.participants(), Some(&[me])
+        )
+        .await?
+        {
+            //#[allow(clippy::type_complexity)]
```
SimonRastikian, line 815: Same here

src/ecdsa/ot_based_ecdsa/triples/generation.rs
```diff
@@ -1010,22 +987,21 @@ async fn do_generation_many<const N: usize>(
...
+    for (from, (their_hat_big_c_i_points, their_phi_proofs)) in
+        recv_from_many::<(Vec<CoefficientCommitment>, Vec<dlog::Proof<C>>)>(
+            &mut chan,
+            wait5,
+            participants.participants(),
+            Some(&[me]),
+        )
+        .await?
+    {
+        //#[allow(clippy::type_complexity)]
```
SimonRastikian, line 1004: Same

src/ecdsa/ot_based_ecdsa/presign.rs
```diff
@@ -109,20 +110,16 @@ async fn do_presign(
     // Receive ej and compute e = SUM_j ej
     // Spec 1.3
     let mut e = e_i;
-    let mut seen = ParticipantCounter::new(&participants);
-    seen.put(me);
-    while !seen.full() {
-        let (from, e_j): (_, Scalar) = chan.recv(wait0).await?;
 
+    for (_from, e_j) in
```
SimonRastikian, line 114: Why _from, can we use _ instead?

src/ecdsa/ot_based_ecdsa/presign.rs
```diff
@@ -152,13 +149,15 @@ async fn do_presign(
...
     let mut beta = beta_i;
-    seen.clear();
-    seen.put(me);
-    while !seen.full() {
-        let (from, (alpha_j, beta_j)): (_, (Scalar, Scalar)) = chan.recv(wait1).await?;
-        if !seen.put(from) {
-            continue;
-        }
+
+    for (_from, (alpha_j, beta_j)) in recv_from_many::<(Scalar, Scalar)>(
```
SimonRastikian, line 153: _from why not _?

src/ecdsa/ot_based_ecdsa/sign.rs
```diff
@@ -98,14 +99,9 @@ async fn do_sign_coordinator(
...
-    let mut seen = ParticipantCounter::new(&participants);
     let mut s = s_i;
-    seen.put(me);
-    while !seen.full() {
-        let (from, s_j): (_, Scalar) = chan.recv(wait0).await?;
-        if !seen.put(from) {
-            continue;
-        }
+    let other_participants = participants.others(me).collect::<Vec<Participant>>();
+    for (_, s_j) in recv_from_many::<Scalar>(&mut chan, wait0, &other_participants, None).await? {
```
SimonRastikian, line 104: Maybe the solution proposed below would make things easier?

src/protocol/helpers.rs
```diff
@@ -0,0 +1,48 @@
...
+) -> Result<T, ProtocolError>
+where
+    T: serde::de::DeserializeOwned,
+{
+    loop {
+        match chan.recv(wait).await? {
+            (p, msg) if p == from => return Ok(msg),
+            _ => {}
+        }
+    }
```
SimonRastikian, line 18: I do not see where this function is used?

src/protocol/helpers.rs
0xsecaas: How about this:
If we use `IntoIterator<Item = Participant>,`:
```rust
pub async fn recv_from_others<T, P>(
    chan: &mut SharedChannel,
    wait: u64,
    participants: P,
) -> Result<Vec<(Participant, T)>, ProtocolError>
where
    T: serde::de::DeserializeOwned,
    P: IntoIterator<Item = Participant>,
{
```
then at call site we could simply pass `participants.others(me)` without ever needing to pass `me` down.

src/ecdsa/ot_based_ecdsa/sign.rs
0xsecaas: Thanks to you, now it looks like this:
```rust
for (_, s_j) in recv_from_others::<Scalar, _>(&mut chan, wait0, participants.others(me)).await?
```

src/protocol/helpers.rs
```diff
@@ -0,0 +1,29 @@
...
+pub async fn recv_from_others<T, P>(
+    chan: &mut SharedChannel,
+    wait: u64,
+    participants: P,
+) -> Result<Vec<(Participant, T)>, ProtocolError>
+where
+    T: serde::de::DeserializeOwned,
+    P: IntoIterator<Item = Participant>,
+{
+    let mut pending: std::collections::HashSet<Participant> = participants.into_iter().collect();
```
0xsecaas, line 17: @SimonRastikian if the number of participants is small (e.g., <16), a small `Vec` search could be even faster due to cache locality than the HashSet.

0xsecaas:
> @0xsecaas could you solve the conflict with main, and fix the clippy issues?

I think its not a blocker to merge, because it says **Changes can be cleanly merged.**

image
---

@gilcu3  Can you please elaborate on `clippy issues`?

netrome (changes requested):
While this generally looks alright, this also introduces timeouts which we shouldn't have in the threshold-signatures library. I don't see what the benefit of multiple intermediate timeouts are here as opposed to letting the caller define their own timeouts.

So please remove these first.

src/protocol/helpers.rs
```diff
@@ -0,0 +1,29 @@
...
+    let mut messages = Vec::with_capacity(pending.len());
+
+    while !pending.is_empty() {
+        let (from, msg) = chan.recv(wait).await?;
+        // extra messages are silently ignored
+        if pending.remove(&from) {
+            messages.push((from, msg));
+        }
```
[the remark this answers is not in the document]
gilcu3: The objective of this PR is to refactor, not to change functionality. So, basically you should try to do exactly the same as before, unless you find a bug that needs fixing. About messages then yes, extra messages were being ignored before, IIRC

src/protocol/helpers.rs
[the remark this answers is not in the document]
gilcu3: The timeout mechanism is not needed here, and this was a conscious design decision. We are taking care of that in other parts of our system, and users of the library should do the same. Therefore, please remove this change as already mentioned [below](https://github.com/near/threshold-signatures/pull/129#pullrequestreview-3297973947)

src/protocol/mod.rs
```diff
@@ -7,6 +7,7 @@
 //! to serialize the emssages it produces.
 pub mod echo_broadcast;
 pub mod errors;
+pub mod helpers;
```
gilcu3, line 10:
```suggestion
pub(crate) mod helpers;
```
no need to expose this outside the crate

src/protocol/helpers.rs
```diff
@@ -0,0 +1,113 @@
...
+use super::{internal::SharedChannel, Participant, ProtocolError};
+
+/// Gather exactly one message from each participant in a group before proceeding.
+///
+/// Note: Result vector order depends on the order messages arrive.
+/// @dev If you ever need deterministic ordering (matching participant list), consider `BTreeMap`.
+pub async fn recv_from_others<T, P>(
+    chan: &SharedChannel,
+    waitpoint: u64,
+    participants: P,
```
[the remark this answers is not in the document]
0xsecaas: Thank you Simon for your patience, please check now.

src/protocol/helpers.rs
0xsecaas: `HashSet` is slightly faster, but the `ParticipantMap` is better for the protocol's overall correctness and 
  predictability.

netrome:
> Yeah it is not because it is a blocker, it is just to be able to run CI (that right now is not possible for some github reason), and to prevent clippy failures. Now that we have so strict rules it is hard to predict.

CI has to be explicitly approved by someone on the team to run it. Don't you have access to approve it and run it?
image

gilcu3:
> > Yeah it is not because it is a blocker, it is just to be able to run CI (that right now is not possible for some github reason), and to prevent clippy failures. Now that we have so strict rules it is hard to predict.
> 
> CI has to be explicitly approved by someone on the team to run it. Don't you have access to approve it and run it? image

Yes I have used it many times already. In this case for some reason it was not possible, probably related with a conflict with main, or a github glitch

src/protocol/helpers.rs
0xsecaas: @SimonRastikian I'm trying to use `ParticipantMap` but I get stuck at converting it to proper iterator, can you help me with this error:

`mismatched types
expected struct `Vec<(protocol::Participant, T)>`
   found struct `participants::ParticipantMap<'_, std::option::Option<T>>``



```rust
pub async fn recv_from_others<T>(
    chan: &SharedChannel,
    waitpoint: u64,
    participants: &ParticipantList,
    me: Participant,
) -> Result<Vec<(Participant, T)>, ProtocolError>
where
    T: serde::de::DeserializeOwned,
{
    let mut seen = ParticipantMap::new(participants);
    seen.put(me, None::<T>);

    while !seen.full() {
        let (from, msg) = chan.recv(waitpoint).await?;
        seen.put(from, msg);
    }

    Ok(seen)
}
```

src/participants.rs
```diff
@@ -269,6 +269,7 @@ impl<'a> ParticipantCounter<'a> {
         inserted
     }
 
+    #[allow(dead_code)]
     /// Clear the contents of this counter.
     pub fn clear(&mut self) {
         for x in &mut self.seen {
```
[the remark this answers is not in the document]
gilcu3: +1

src/ecdsa/ot_based_ecdsa/triples/generation.rs
```diff
@@ -510,6 +482,16 @@ async fn do_generation_many<const N: usize>(
...
+    #[allow(clippy::struct_field_names)]
+    struct ParallelToMultiplicationTaskOutput {
+        big_e_v: Vec<PolynomialCommitment>,
+        big_f_v: Vec<PolynomialCommitment>,
+        big_l_v: Vec<PolynomialCommitment>,
+        big_c_v: Vec<ProjectivePoint>,
+        a_i_v: Vec<Scalar>,
+        b_i_v: Vec<Scalar>,
+    }
+
```
gilcu3, line 494: same here, please don't move it around

src/ecdsa/ot_based_ecdsa/triples/generation.rs
```diff
@@ -611,16 +593,6 @@ async fn do_generation_many<const N: usize>(
         )
     };
 
-    #[allow(clippy::items_after_statements)]
-    struct ParallelToMultiplicationTaskOutput<'a> {
-        seen: ParticipantCounter<'a>,
```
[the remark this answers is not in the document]
0xsecaas: `seen` was used primarily for the while loop:
```rust
    while !seen.full() {
```
which is not refactored into a centralized function called `recv_from_others`

src/ecdsa/ot_based_ecdsa/triples/generation.rs
0xsecaas: Sure thing.
I suggest suffixing the second one with `Many` to make it explicitly clear it contains many items. ok @gilcu3 ?

gilcu3: Yes, that would be helpful. The current name is a bit misleading given the presence of the other one

PR #131, opened 2025-10-02
preliminary doc for the communication channels

### LINKED ISSUE #98: Document the communication channels
Currently, we do not have any documentation about how the protocols work on a networking level.
What is currently missing is a proper Readme describing our underlying assumptions of the communication channels and the different communication functions, i.e., send_many, send_private, and broadcast.

All of these assume authenticated channel but only send_private assumes also encrypted channel. In practice, we provide both authentication and confidentiality by running the protocols over TLS.
We also need to define the notation used in the documentation of ECDSA for sending and receiving, namely the triangle and circle, etc...

Acceptance Criteria:
* Good quality documentation
* Proper explanation of the different communication channels types
* Definition of the notation used in other docs for sending and receiving elements.

0xsecaas:
closes #98

### DIFF
--- a/docs/network_layer.md
+++ b/docs/network_layer.md
@@ -34,12 +34,7 @@ The protocol implementation provides several communication primitives:
 
   - **Totality**: If some message is delivered by any correct process, then every correct process eventually delivers some message.
 
-These guarantees hold under standard _threshold assumptions_ (where `n` = _total participants_, `f` = _maximum faulty nodes tolerated_):
-   - **DKG**: Threshold `t = f + 1`. Requires `f ≤ ⌊N/3⌋`. Example: with `n = 7`, `f = 2` -> `t = 3`.
-
-   - **OtBasedECDSA**: Threshold `t = f + 1`. No additional requirement. Example: with `n = 7`, `f = 2` -> `t = 3`.
-   
-   - **Robust ECDSA**: Threshold `t = f`. Requires `2f + 1 ≤ N`. Example: with `n = 7`, `f = 2` -> `t = 2`.
+> To guarantee the security notions given by the Byzantine Reliable Broadcast, we assume that `N >= 3f +1`. This bound originates from the classical Byzantine fault tolerance model [LSP82](https://lamport.azurewebsites.net/pubs/byz.pdf), which ensures both safety and liveness under such faults.
 
 ## Documentation Notation
 In protocol specifications (particularly for ECDSA), we use the following symbols to describe actions:


src/protocol/README.md
```diff
@@ -0,0 +1,37 @@
...
+## Core Assumptions
+
+Our protocols operate on two fundamental assumptions about the network channels:
+
+- **Authenticated Channels:** All messages are sent over authenticated channels. Senders' identities are always verifiable.
+- **Confidentiality for Private Messages:** Channels used for private messages (`send_private`) must be encrypted.
+
+<details>
+  <summary>Practical Implementation</summary>
+  In practice, we satisfy both requirements by running all protocols over a network where each participant is connected via a TLS channel with a certified key pair. This ensures all communication is both authenticated and encrypted.
```
SimonRastikian, line 14: I would say:
... where participants are peer-to-peer connected via a TLS channel ... 

... This ensures both, authentication and confidentiality.

src/protocol/README.md
```diff
@@ -0,0 +1,37 @@
...
+</details>
+
+## Communication Primitives
+
+The protocol implementation provides several communication primitives:
+
+- **`send_many`**: Sends a message to all other participants. This is a basic, non-blocking broadcast that does not guarantee message delivery or ordering. It's suitable for rounds where eventual delivery is sufficient.
```
SimonRastikian, line 21: I would say:
... Sends a message to all participants except the sender itself...

src/protocol/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+- **`send_private`**: Sends a message to a single, specific participant. The underlying channel is assumed to be confidential.
+
+- **Reliable Broadcast (`echo_broadcast`)**: A higher-level protocol that ensures all honest participants agree on the same set of messages from senders. It is built using `send_many` and guarantees that a message is delivered if and only if all honest parties receive it. This is critical for rounds requiring consensus.
```
SimonRastikian, line 25: I would say:
... A complex protocol ... 
... agree on the same message...

I would not say it is build using send_many because it confuses people.  
Please mention that the protocol is the Byzantine Reliable Broadcast protocol
I would develop about the guarantees:
* Validity: If a correct process p broadcasts a message m, then p eventually delivers m.
* No duplication: No message is delivered more than once.
* No creation: If a process delivers a message m with sender s, then m was previously broadcast by process s.
* Agreement: If a message m is delivered by some correct process, then m is eventually delivered by every correct process.
* Totality: If some message is delivered by any correct process, every correct process eventually delivers a message


Please add info about the threshold assumptions.

SimonRastikian (commented):
Please address the changes before I merge :) 
Thank you for your contribution

gilcu3:
@0xsecaas thank you for taking care of this! Could you move this doc to the docs/ folder? A filename like `network.md` or `network_layer.md` would make sense there.

src/protocol/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Documentation Notation
+
+In our protocol specifications (particularly for ECDSA), we use the following symbols to describe actions:
+
+| Symbol | Meaning | Description |
+| :---: | :--- | :--- |
+| `⋆` | **Send** | A participant sends a message to one or more others. |
+| `⚫` | **Receive** | A participant waits to receive a message. |
+| `🔺` | **Assert** | A participant makes an assertion. The protocol aborts if it fails. |
```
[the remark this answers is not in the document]
0xsecaas: @gilcu3 done.

@SimonRastikian `docs/ecdsa/ot_based_ecdsa/triples.md` and `docs/ecdsa/robust_ecdsa/signing.md` use the red color for private send.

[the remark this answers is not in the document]
0xsecaas: @gilcu3 So no `red star` for `private send` anymore? and all `send` (public/private) are represented by a uniform `star`(\star) ?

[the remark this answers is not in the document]
0xsecaas: In that case, we don't need to change anything AFAIK.

docs/network_layer.md
```diff
@@ -0,0 +1,52 @@
...
+  - **Agreement**: If a message `m` is delivered by some correct process, then `m` is eventually delivered by every correct process.
+
+  - **Totality**: If some message is delivered by any correct process, then every correct process eventually delivers some message.
+
+These guarantees hold under standard _threshold assumptions_ (where `n` = _total participants_, `f` = _maximum faulty nodes tolerated_):
+   - **DKG**: Threshold `t = f + 1`. Requires `f ≤ ⌊N/3⌋`. Example: with `n = 7`, `f = 2` -> `t = 3`.
+
+   - **OtBasedECDSA**: Threshold `t = f + 1`. No additional requirement. Example: with `n = 7`, `f = 2` -> `t = 3`.
+   
+   - **Robust ECDSA**: Threshold `t = f`. Requires `2f + 1 ≤ N`. Example: with `n = 7`, `f = 2` -> `t = 2`.
```
SimonRastikian, line 42: Let's not talk about these things here. 
Also FYI: a Distributed Systems "threshold" for async systems is less than 1/3. But a Cryptographic "threshold" for signature schemes really depends on the scheme. 

One can assume 1/3 parties to be faulty in a distributed system but they are generating key shares with a cryptographic threshold of say n-1.
I know this seems weird but say you assume when running a DKG (with echo-broadcast) only 1/3 are malicious. But the moment you sign, then up to n-1 could be malicious and your protocol remains secure.

docs/network_layer.md
```diff
@@ -0,0 +1,52 @@
...
+  - **No duplication**: No message is delivered more than once.
+  
+  - **No creation**: If a process delivers a message `m` with sender `s`, then m was previously broadcast by process `s`.
+
+  - **Agreement**: If a message `m` is delivered by some correct process, then `m` is eventually delivered by every correct process.
+
+  - **Totality**: If some message is delivered by any correct process, then every correct process eventually delivers some message.
+
+These guarantees hold under standard _threshold assumptions_ (where `n` = _total participants_, `f` = _maximum faulty nodes tolerated_):
+   - **DKG**: Threshold `t = f + 1`. Requires `f ≤ ⌊N/3⌋`. Example: with `n = 7`, `f = 2` -> `t = 3`.
```
SimonRastikian, line 38: I would simply say something like:
To guarantee the security notions given by the Byzantine Reliable Broadcast, we assume that N>= 3f +1. This bound originates from the classical Byzantine fault tolerance model [[LSP82](https://lamport.azurewebsites.net/pubs/byz.pdf)], which ensures both safety and liveness under such faults.

SimonRastikian (commented):
All good, just some small notes. Thanks :)

docs/network_layer.md
0xsecaas: Two separate phases, two different thresholds. understood.
Thanks for clarifying.

SimonRastikian (approved):

PR #172, opened 2025-10-25
fix: remove unsafe code

### LINKED ISSUE #150: [Sec] Potential Endianness Vulnerability (Affects Big-Endian Systems)
### Background

There appears to be a **latent endianness dependency** in the `threshold-signatures` cryptographic state handling.
On **big-endian architectures**, internal Keccak state serialization may be interpreted inconsistently, leading to **state corruption** or **forked signatures** across heterogeneous networks.
This can result in **network divergence**, **invalid signature verification**, or other **security-critical inconsistencies** when nodes of different architectures interoperate.

### User Story

I want to confirm whether cryptographic serialization and hashing logic are endian-agnostic or implicitly little-endian, so that the library remains secure and behaves deterministically across all supported architectures.

### Acceptance Criteria

- The codebase is reviewed for **endianness-dependent logic** in Keccak and serialization layers.
- A **test or build check** confirms consistent behavior between little-endian and big-endian environments.
- If portability is not a design goal: **a compile-time assertion** restricts builds to little-endian only.
- If portability is desired: a **portable**, **endian-agnostic serialization** mechanism is implemented.
- Documentation reflects the decision and explicitly states supported architectures.
- Public confirmation that the endianness concern is either **validated** or **ruled out**.

# Conceptual Model
```mermaid
graph TD
A[threshold-signatures] -->|Portability Goal| B[Supports Big-Endian?]
B --> |No| G[Implicit Little-Endian Dependency]
B --> |Yes| F[Potential Vulnerability]

G -->|Fix| I[Solution 1: Assert Little-Endian Only]
F -->|Fix| H[Solution 2: Implement Endian-Agnostic Serialization]
```

gilcu3:
Closes #150

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,9 @@ jobs:
           cargo fmt -- --check
 
       - name: Install cargo-nextest
-        run: cargo install cargo-nextest
+        uses: taiki-e/install-action@v2
+        with:
+          tool: nextest@0.9.106
 
       - name: Run cargo-nextest
         run: cargo nextest run --locked --all-features --cargo-profile test-release
@@ -63,7 +65,9 @@ jobs:
       - name: Install cargo-llvm-cov
         uses: taiki-e/install-action@cargo-llvm-cov
       - name: Install cargo-nextest
-        run: cargo install cargo-nextest
+        uses: taiki-e/install-action@v2
+        with:
+          tool: nextest@0.9.106
       - name: Generate code coverage
         run: cargo llvm-cov nextest --cargo-profile test-release --all-features --workspace --codecov --output-path codecov.json
       - name: Upload coverage to Codecov
@@ -72,4 +76,3 @@ jobs:
           token: ${{ secrets.CODECOV_TOKEN }}
           files: codecov.json
           fail_ci_if_error: false
-
--- a/ISSUE_TEMPLATE/config.yml
+++ b/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,2 @@
+blank_issues_enabled: false
+contact_links: []
--- a/ISSUE_TEMPLATE/new_task.yml
+++ b/ISSUE_TEMPLATE/new_task.yml
@@ -1,0 +1,38 @@
+# Overrides https://github.com/near/.github/blob/main/.github/ISSUE_TEMPLATE/new-task.yml
+
+name: New Task
+description: Create a new work task.
+labels: ["issue"]
+body:
+  - type: markdown
+    attributes:
+      value: "## 📝 New Task"
+
+  - type: textarea
+    id: background
+    attributes:
+      label: "Background"
+      placeholder: "Brief description of what this task will accomplish. What is the goal and why should this be done?"
+    validations:
+      required: true
+
+  - type: textarea
+    id: user-story
+    attributes:
+      label: "User Story"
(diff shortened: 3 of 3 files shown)


src/crypto/proofs/strobe.rs
```diff
@@ -1,33 +1,47 @@
...
+    result
 }
 
-/// This is a wrapper around 200-byte buffer that's always 8-byte aligned
-/// to make pointers to it safely convertible to pointers to [u64; 25]
-/// (since u64 words must be 8-byte aligned)
-#[derive(Clone, Zeroize)]
+fn untransmute_state(transmuted_state: [u64; 25], state: &mut KeccakState) {
+    for (i, ti) in transmuted_state.iter().enumerate() {
+        state[8 * i..8 * i + 8].copy_from_slice(&ti.to_le_bytes());
```
SimonRastikian:
```suggestion
        state[i<<3..(i+1)<<3].copy_from_slice(&ti.to_le_bytes());
```

src/crypto/proofs/strobe.rs
```diff
@@ -1,33 +1,47 @@
...
+    }
+}
+
+fn keccak_f1600_wrapper(state: &mut KeccakState) {
+    let mut transmuted_state = transmute_state(state);
+    keccak::f1600(&mut transmuted_state);
+    untransmute_state(transmuted_state, state);
+}
```
SimonRastikian: I do not get what this function does

gilcu3: It is certainly non-trivial, but basically it is strictly equivalent to what we had before:
```rust
keccak::f1600(transmute_state(&mut st));
```

The new `transmute_state` function does exactly the same, but to remain safe it creates a copy of the data. Then that data is passed as a mutable reference to `keccak::f1600`, which modifies its value. As we need the modified value in the original form, we call `unstransmute_state` to revert the first operation applied. 

If you want to check why is that last step needed, comment that line and you will see some tests failing.

0xsecaas:
> > @gilcu3 Why not just [manually copy](https://github.com/near/threshold-signatures/pull/173) data?
> 
> Manually copying the data is basically what I am doing in this PR, isn't it?

Not quite — #172 adds an external crate for what’s already achievable with plain byte copies.
#173 does the same safely using only the standard library, no extra dependencies.

gilcu3:
> > > @gilcu3 Why not just [manually copy](https://github.com/near/threshold-signatures/pull/173) data?
> > 
> > 
> > Manually copying the data is basically what I am doing in this PR, isn't it?
> 
> Not quite — #172 adds an external crate for what’s already achievable with plain byte copies. #173 does the same safely using only the standard library, no extra dependencies.

If you meant only about the extra dependency, in this case it is one we are already use in other repos and are very comfortable with, which basically helps making code more ergonomic while avoiding having boilerplate code.

src/crypto/proofs/strobe.rs
SimonRastikian: I think I understood, the `keccak:f1600` performs the permutation then the state needs to be untransmuted back to le bytes

SimonRastikian (approved):
Approved. 
Feel free to merge but, please either open an issue regarding making the code more ergonomic using deref-more everywhere or push a new commit that does away with the crate.

gilcu3:
> Approved. Feel free to merge but, please either open an issue regarding making the code more ergonomic using deref-more everywhere or push a new commit that does away with the crate.

Makes sense, issue created https://github.com/near/threshold-signatures/issues/175

PR #179, opened 2025-10-27
fix: add proper checks in BLS signature verification

### LINKED ISSUE #157: BLS signature verification
### Background

Even though we are currently only using the BLS signature verification in tests, we should make sure it is standards compliant.

### User Story

Implementation should be hard to misuse

### Acceptance Criteria

The BLS signature verification is standard compliant.

_No response_

gilcu3:
Closes #157

### DIFF
--- a/src/confidential_key_derivation/ciphersuite.rs
+++ b/src/confidential_key_derivation/ciphersuite.rs
@@ -214,6 +214,8 @@ impl frost_core::Group for BLS12381G1Group {
     }
 }
 
+/// BLS signature verification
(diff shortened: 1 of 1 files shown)


src/confidential_key_derivation/ciphersuite.rs
```diff
@@ -217,13 +219,21 @@ pub fn verify_signature(
     msg: &[u8],
     signature: &Signature,
 ) -> Result<(), frost_core::Error<BLS12381SHA256>> {
+    let element1: G1Affine = signature.into();
+    if (!element1.is_on_curve() | !element1.is_torsion_free() | element1.is_identity()).into() {
```
SimonRastikian: Why are those or not and?
Why do you want that element1 be the identity?

PR #180, opened 2025-10-27
Simon/code refactor rerandomization

### LINKED ISSUE #165: Audit: Add the tweak in the rerandomization HKDF function
GS21 mentions heuristical approach to generate the rerandomization string rho for ECDSA scheme and prevent Wagner attack.
Our current version does not include the tweak (key derivation computed scalar) and the auditors believe it is good to integrate it in the HKDF function that generates the rerandomization string.

Acceptance Criteria:
* Modify the code to add the tweak in the rerandomization generation.

SimonRastikian:
Closes #165.

### DIFF
--- a/src/ecdsa/mod.rs
+++ b/src/ecdsa/mod.rs
@@ -243,7 +243,6 @@ mod test {
     }
 
     #[test]
-    #[allow(non_snake_case)]
     fn keygen_output_should_be_serializable() {
         // Given
         let mut rng = MockCryptoRng::new([1; 8]);


src/ecdsa/mod.rs
```diff
@@ -90,6 +90,7 @@ pub type SignatureOption = Option<Signature>;
 /// be public, freshly generated, and unpredictable.
 pub struct RerandomizationArguments {
     pub pk: AffinePoint,
```
gilcu3: we need to add a comment here mentioning that this is now the MPC network public key, and not the derived public key

gilcu3 (commented):
I left a few comments that might or might not need a few changes.

The only minor blocker is that we need to update the docs for this change.

src/ecdsa/mod.rs
```diff
@@ -238,7 +243,7 @@ mod test {
 
     #[test]
     #[allow(non_snake_case)]
-    fn keygen_output__should_be_serializable() {
+    fn keygen_output_should_be_serializable() {
```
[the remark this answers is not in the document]
gilcu3: I think we should leave it as it was, even without a convention. IMHO unless previous code is wrong, status quo takes precedence.

src/ecdsa/mod.rs
SimonRastikian: It does not matter. Both work well and I prefer not to enforce anything on the library user

gilcu3: Yes, I think that is technically correct, but we wanted to align with the paper as much as possible, so I think we should simply use the original public key. I don't think that would restrict a library user in any way. The comment would not enforce it anyway, as the type is the same, but hint that we expected it to be used as in the paper.

Issue #185, opened 2025-10-29
[SEC] Remote DoS -- Node OOM crash


0xsecaas:
### Background

# Unbounded message buffer allows remote Out Of Memory (async waitpoint flood)

## Analysis

  - The async inbox keeps a `HashMap<MessageHeader, SubMessageQueue>` and [blindly allocates](https://github.com/near/threshold-signatures/blob/main/src/protocol/internal.rs#245) a new [unbounded queue](https://github.com/near/threshold-signatures/blob/main/src/protocol/internal.rs#200) for every header it sees.
  - a malicious peer can fabricate arbitrary _channel_/_waitpoint_ identifiers that the honest state machine will never [poll](https://github.com/near/threshold-signatures/blob/main/src/protocol/internal.rs#295).
  - Those spoofed messages are never drained, so the map and the buffered payloads grow without bound; repeated packets trigger memory exhaustion and **take the participant down**.

### Quick patch
Treat the buffer as a cache—only accept a bounded number of outstanding waitpoints per channel and evict or drop messages once the cap is hit. That keeps the memory footprint under control even if an attacker keeps injecting junk, at the cost of **possible message loss under attack**.

### Robust fix
introduce explicit waitpoint registration/capabilities. Require the receiver to announce which waitpoints are valid (or hand out per-waitpoint tokens) before accepting traffic. Messages that don’t match an outstanding registration are rejected up front, which both throttles memory use and ensures the state machine only receives traffic it expects.

SimonRastikian:
Seems like the quick patch is not a real patch: The attacker can still fill the number of outstanding waitpoints per channel which will lead the honest participant to drop potentially honest messages. Do you agree?

PR #186, opened 2025-10-29
POC demonstrating waitpoint flood DoS

0xsecaas:
# Proof of Concept

This unit test demonstrates the attack (#185) by injecting 512 fake waitpoints and [asserting](https://github.com/near/threshold-signatures/blob/main/src/protocol/internal.rs#538) that every one of them sticks around.
 
Reproduce with `cargo test attacker_can_fill_message_buffer_with_unused_waitpoints`.

--- 
⛑️  Let me know if you want help prototyping a fix or adding guardrails to the message dispatch layer.

---

### DIFF
--- a/src/protocol/internal.rs
+++ b/src/protocol/internal.rs
@@ -528,3 +528,35 @@ pub fn make_protocol<T: Send>(
 ) -> impl Protocol<Output = T> {
     ProtocolExecutor::new(comms, fut)
 }
+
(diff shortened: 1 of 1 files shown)


SimonRastikian:
I approve this PR. Let's see how we can fundamentally improve the design even though this attack is only produced by one of the signers and not by any external malicious party.

PR #193, opened 2025-10-31
Simon/naive ecdsa bench

### LINKED ISSUE #177: [Benchmarking] Criterion Benchmarks for ECDSA schemes
Benchmarking ECDSA schemes (Robust ECDSA and OT-based ECDSA) in a proper manner might take a lot of time to be completed.
An alternative quick solution would be to utilize criterion and perform a naive benchmarking.
The benchmark would allow fixing a threshold, and then running the multiple protocols (side-by-side) to be able to compare their performance.

Note that we do realize that this benchmarking is naive for several reasons:
1) It runs multiple participants in a sequential manner which prevents us from clearly analyze the difference between the schemes especially
that they require different numbers of signers for the same threshold
2) The benchmarking does not take into consideration the parallelization that is happening in the triple generation of OT-based ECDSA which means it could be slower than running it in reality.
3) The communication between the participants happen in a sequential manner, this means that a participant sending one message to all would make n send requests while the other participants are actually waiting
4) Due to the sequentiality in the communication, it is difficult to represent latency of the communication layer 

This Benchmark even though not perfect, gives us an idea about how the time required when running the schemes.

Acceptance Criteria:
* Test that the signature generated are indeed valid and that no participant aborted during the benchmarking process.
* When running the benchmarking, allow taking some parameters as inputs such as (threshold, presign, sign, number_of_rounds, all)
that specifies correspondingly the threshold, the algorithms to run and how many times those latter should be ran.
* Specify in a comment why are these benchmarking naive
* Give a small summary of the numbers found in a readme and for multiple thresholds

Reference: #152.

SimonRastikian:
Closes #177

### DIFF
--- a/benches/naive_ot_based_ecdsa.rs
+++ b/benches/naive_ot_based_ecdsa.rs
@@ -39,7 +39,7 @@ fn bench_triples(c: &mut Criterion) {
     group.measurement_time(std::time::Duration::from_secs(200));
 
     group.bench_function(
-        format!("ot_ecdsa_triples_naive_MAX_MALICIOUS{max_malicious}_PARTICIPANTS_{num}"),
+        format!("ot_ecdsa_triples_naive_MAX_MALICIOUS_{max_malicious}_PARTICIPANTS_{num}"),
         |b| {
             b.iter_batched(
                 || prepare_triples(participants_num()),
@@ -61,7 +61,7 @@ fn bench_presign(c: &mut Criterion) {
     let two_triples = run_protocol(protocols).expect("Running triple preparations should succeed");
 
     group.bench_function(
-        format!("ot_ecdsa_presign_naive_MAX_MALICIOUS{max_malicious}_PARTICIPANTS_{num}"),
+        format!("ot_ecdsa_presign_naive_MAX_MALICIOUS_{max_malicious}_PARTICIPANTS_{num}"),
         |b| {
             b.iter_batched(
                 || prepare_presign(&two_triples),
@@ -88,7 +88,7 @@ fn bench_sign(c: &mut Criterion) {
     result.sort_by_key(|(p, _)| *p);
 
     group.bench_function(
-        format!("ot_ecdsa_sign_naive_MAX_MALICIOUS{max_malicious}_PARTICIPANTS_{num}"),
+        format!("ot_ecdsa_sign_naive_MAX_MALICIOUS_{max_malicious}_PARTICIPANTS_{num}"),
         |b| {
             b.iter_batched(
                 || prepare_sign(&result, pk),
--- a/benches/naive_robust_ecdsa.rs
+++ b/benches/naive_robust_ecdsa.rs
@@ -31,7 +31,7 @@ fn bench_presign(c: &mut Criterion) {
     let mut group = c.benchmark_group("presign");
     group.measurement_time(std::time::Duration::from_secs(300));
     group.bench_function(
-        format!("robust_ecdsa_presign_naive_MAX_MALICIOUS{max_malicious}_PARTICIPANTS_{num}"),
+        format!("robust_ecdsa_presign_naive_MAX_MALICIOUS_{max_malicious}_PARTICIPANTS_{num}"),
         |b| {
             b.iter_batched(
                 || prepare_presign(participants_num()),
@@ -54,7 +54,7 @@ fn bench_sign(c: &mut Criterion) {
     result.sort_by_key(|(p, _)| *p);
 
     group.bench_function(
-        format!("robust_ecdsa_sign_naive_MAX_MALICIOUS{max_malicious}_PARTICIPANTS_{num}"),
+        format!("robust_ecdsa_sign_naive_MAX_MALICIOUS_{max_malicious}_PARTICIPANTS_{num}"),
         |b| {
             b.iter_batched(
                 || prepare_sign(&result, pk),


benches/protocol_benchmarks/naive_benchmarks/mod.rs
```diff
@@ -0,0 +1,40 @@
...
+use frost_secp256k1::{Secp256K1ScalarField, VerifyingKey};
+use rand_core::{CryptoRngCore, OsRng};
+
+// Outputs pk, R, hash, participants, entropy, randomness
+pub fn generate_rerandpresig_args(
+    rng: &mut impl CryptoRngCore,
+    participants: &[Participant],
+    pk: VerifyingKey,
+    big_r: AffinePoint,
+) -> (RerandomizationArguments, Scalar) {
```
gilcu3: This looks like the perfect function for test-utils, so I would put it there, unless you think it does not make sense to reuse it anywhere else.

benches/utils/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+// fix malicious number of participants
+pub static MAX_MALICIOUS: LazyLock<usize> = std::sync::LazyLock::new(|| {
+    env::var("MAX_MALICIOUS")
+        .ok()
+        .and_then(|v| v.parse().ok())
+        .unwrap_or(6)
+});
+
+// Outputs pk, R, hash, participants, entropy, randomness
+pub fn generate_rerandpresig_args(
```
gilcu3: nit: probably better to add ecdsa to the function name here

benches/naive_robust_ecdsa.rs
```diff
@@ -0,0 +1,158 @@
...
+    2 * *crate::MAX_MALICIOUS + 1
+}
+
+/// Benches the presigning protocol
+fn bench_presign(c: &mut Criterion) {
+    let mut group = c.benchmark_group(format!(
+        "Presign: {} malicious parties and {} participating parties\n",
+        *MAX_MALICIOUS,
+        participants_num()
+    ));
```
gilcu3: same

benches/naive_ot_based_ecdsa.rs
```diff
@@ -0,0 +1,233 @@
...
+    });
+}
+
+/// Benches the presigning protocol
+fn bench_presign(c: &mut Criterion) {
+    let mut group = c.benchmark_group(format!(
+        "Presign: {} malicious parties and {} participating parties\n",
+        *MAX_MALICIOUS,
+        participants_num()
+    ));
```
gilcu3: same

gilcu3 (changes requested):
I think it looks good overall, although I am requesting two changes:

- bench group names should not contain punctuation signs or spaces, so that they can be nicely used with filters from the command line. I marked several of them, but might have missed some. For the same reason bench function names should follow the same pattern.
- currently the benchmarks are using OsRng. We could use a seedable deterministic RNG (the rand crate contains some) so that we can somehow compare results. I know this will not make our make our benches deterministic yet (the non-deterministic order of messages still matters), but would be a step in that direction that can already be taken here with little effort.

benches/naive_ot_based_ecdsa.rs
```diff
@@ -0,0 +1,233 @@
...
+fn prepare_sign(
+    result: &[(Participant, PresignOutput)],
+    pk: VerifyingKey,
+) -> Vec<(Participant, Box<dyn Protocol<Output = SignatureOption>>)> {
+    // collect all participants
+    let participants: Vec<Participant> =
+        result.iter().map(|(participant, _)| *participant).collect();
+
+    // choose a coordinator at random
+    let index = OsRng.gen_range(0..result.len());
```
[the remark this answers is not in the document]
SimonRastikian: Why should we?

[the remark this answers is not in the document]
SimonRastikian: No I do not think this has any impact over performance or any utility as the benches are not tests so you would not want repetition. Even if you do, you will not get the exact same numbers

benches/naive_ot_based_ecdsa.rs
```diff
@@ -0,0 +1,233 @@
...
+            run_protocol,
+            criterion::BatchSize::SmallInput,
+        );
+    });
+}
+
+type PreparedTriples = Vec<(
+    Participant,
+    Box<dyn Protocol<Output = Vec<(TripleShare, TriplePub)>>>,
+)>;
```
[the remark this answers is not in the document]
SimonRastikian: What do you mean? Why would prepare presign take Protocols outputs as its input?

gilcu3:
@SimonRastikian the change in https://github.com/near/threshold-signatures/pull/193/commits/864ec486de7079dd5580845866cff002560255ee does not fix the issue I pointed with function names. Notice when you call something like:
```rust
group.bench_function(description, |b| {
        b.iter_batched(
            || prepare_triples(participants_num()),
            run_protocol,
```
the first parameter (here you called it `description`) is the function name, so it should not have any symbols or spaces, as normal function names

SimonRastikian:
@gilcu3  I did this [3ca7c34](https://github.com/near/threshold-signatures/pull/193/commits/3ca7c34c52296b2bde001ff03e8db86f4f79d236) instead. But since I want to still have a description of how many participants and what is the threshold, I printed out the description. The problem now is that the description will get printed all the time on every cargo bench call e.g. `cargo bench -- presign`. I think this should be fine but still

gilcu3:
Also we still need to enable the `html_report` feature. Just doing that I am already able to get, which is quite nice

image

gilcu3 (approved):
Thank you, left only a nit about missing underscore

PR #197, opened 2025-11-03
feat: alternative bench organization

gilcu3:
A more natural bench organization, [context](https://github.com/near/threshold-signatures/pull/193#discussion_r2480987746)


SimonRastikian:
Please could you add a description?

gilcu3:
> Please could you add a description?

I added explanation and context in https://github.com/near/threshold-signatures/pull/193#discussion_r2485704806

SimonRastikian (commented):
I couple of notes:
1) I would keep the crypto files separate as we might add more and more functionalities as the code grows (e.g. multi-scalar mult)
2) I insist on keeping naive_benchmarks in the files names as these are not the more advanced ones we use.
3) I could not find the help function which prints how to run the benches. I would keep it as the user cannot guess that the MAX_PARTICIPANT could be set as an env variable

gilcu3:
> Looks good to me thanks!

Feel free to merge then, the original PR is still yours :) I will then continue the review there ASAP

PR #203, opened 2025-11-04
Added complete threshold-signatures Rust implementation

### LINKED ISSUE #200: Audit: Missing zeroization of presignature
### Background

The PresignOutput and RerandomizedPresignOutput structures contain sensitive 
cryptographic material that remains in memory after use because they do not implement 
automatic memory zeroization. These structures hold secret presignature shares that, if 
reused across multiple signatures, allow an attacker to recover the private key share 
through standard ECDSA nonce reuse attacks. 
The PresignOutput and RerandomizedPresignOutput structures are defined without zeroization.

Although the zeroize crate (version 1.8.1) is available in the project dependencies, neither 
structure derives or implements the Zeroize or ZeroizeOnDrop traits. The underlying 
Scalar type from k256 implements DefaultIsZeroes but does not automatically zeroize 
on drop. Without explicit zeroization, presignature data persists in memory until the pages 
are reused, potentially surviving for extended periods in heap memory.

### User Story

The exploit scenario:
An application uses the robust ECDSA implementation to sign transactions. A memory 
disclosure vulnerability exists elsewhere in the application, allowing an attacker to read 
portions of the process memory. After a signing operation completes, the presignature 
values remain in memory because they are not zeroized. The attacker exploits the memory 
disclosure to extract a participant's presignature data, specifically the c and beta values. 
The attacker can recover the victim's private key share x since  beta = c . x

### Acceptance Criteria

Short term, implement automatic memory zeroization for both presignature structures by 
deriving the ZeroizeOnDrop trait. 

Long term, consider wrapping presignature data in a secure container type that enforces 
single-use semantics through the type system, making it impossible to reuse a presignature 
even if the zeroization is bypassed. Additionally, conduct a comprehensive audit of all 
cryptographic material in the codebase to ensure consistent application of zeroization 
patterns, and establish coding standards that require ZeroizeOnDrop for all structures 
containing secret values.

_No response_

yummyash:
Closes #200

### DIFF
--- a/src/eddsa/mod.rs
+++ b/src/eddsa/mod.rs
@@ -1,20 +1,92 @@
-//! This module serves as a wrapper for Ed25519 scheme.
-pub mod sign;
-#[cfg(test)]
-mod test;
+// src/ecdsa/mod.rs
+//! ECDSA module including presignature and rerandomization logic.
 
-use crate::crypto::ciphersuite::{BytesOrder, Ciphersuite, ScalarSerializationFormat};
-pub use frost_ed25519::Ed25519Sha512;
+mod secure_scalar;
+use secure_scalar::SecureScalar;
 
-pub type KeygenOutput = crate::KeygenOutput<Ed25519Sha512>;
+use k256::Scalar;
+use zeroize::{Zeroize, ZeroizeOnDrop};
 
-impl ScalarSerializationFormat for Ed25519Sha512 {
-    fn bytes_order() -> BytesOrder {
-        BytesOrder::LittleEndian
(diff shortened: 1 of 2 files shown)


SimonRastikian:
Thanks for your contribution. Since the PR does not actually solve the original problem #200, I will then close it to keep the record clean. I suggest you read carefully #200 and tackle the ECDSA scheme instead by allowing zeroization on the presignatures as mentioned in the issue.
If this is too complex but you would still like to contribute to the project, I suggest you pick another issue tagged with "good first issue". These are supposed to be easier to handle.

PR #205, opened 2025-11-05
fix: PolynomialCommitment serialization

### LINKED ISSUE #162: Audit: PolynomialCommitment deserialization does not trim the coefficient list
PolynomialCommitment::deserialize function allows trailing identity elements in  its representation, causing functions that rely on a trimmed commitment list, such as the degree function, to return incorrect results. This could be problematic in the OT-based ECDSA implementation if the serialized data could contain the identity point, (but this is not the case). 

Acceptance Criteria:
* Use the PolynomialCommitment::new function in the deserialization routine instead of building it from scratch.  This ensures that all created elements of that type enforce the same invariants.

gilcu3:
Closes #162

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,7 +56,7 @@ test-utils = ["rand"]
 [dev-dependencies]
 criterion = "0.7.0"
 bincode = { version = "2.0.1", features = ["serde"] }
-rand = "0.8.5"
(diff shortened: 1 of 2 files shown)


SimonRastikian (approved):
Nice! Please modify the todo list in the private repo

PR #209, opened 2025-11-06
chore: add action to add issues to project automatically

### LINKED ISSUE #202: Github Yaml
### Background

Github currently does not switch an issue to in-review mode on the our kanban board which makes things a little annoying to have to do this manually.
Extra: Currently opening an issue would not automatically send it as a backlog to the Kanban project. It would be happy moments if it does so.

### User Story

_No response_

### Acceptance Criteria

Modify the yaml file to add automatic in-review mode when opening a PR
[Optional] Each freshly opened issue should be automatically assigned to the MPC Team Kanban project as a Backlog

_No response_

gilcu3:
Closes #202 

- Notice we cannot test this until after it is merged (but I tested in another repo)
- I am using a token I generated, it would be better to use one from the organization. The change does not affect the PR, just the config in github
- This cannot verify who opened the issue. If we face any kind of DoS, we will need to disable it

### DIFF
--- a/.github/workflows/add-new-issues-to-project.yml
+++ b/.github/workflows/add-new-issues-to-project.yml
@@ -9,6 +9,8 @@ jobs:
   add-to-project:
     name: Add issue to project
     runs-on: ubuntu-latest
+    permissions:
+        issues: read
     steps:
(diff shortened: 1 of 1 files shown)


SimonRastikian (approved):
I will approve this and trust your construction.

PR #214, opened 2025-11-11
Changing fixed outputs

### LINKED ISSUE #107: Deterministic unit tests
### Background

We should have deterministic tests in our code base.

### User Story

Changes in our crypto dependencies and implementation should be noticeable in tests, so that changes happen knowingly.

### Acceptance Criteria

We have deterministic tests. This includes creating a better MockCrypto implementation.

_No response_

### LINKED ISSUE #213: Improve the MockCryptoRng
### Background

Currently the implementation of MockCryptoRng is extremely basic. It takes a 64 bit input and then outputs it over and over again as an output. 
This  is enough for most of test cases but things start to become unclear when wanting to compare cloned Rngs used for running snapshot multiple times.
An improvement of this Rng is to make it hash based. This would keep the determinism and clonability properties but also allows to have more variable output bytes which helps comparing cloned rng outputs together.

### User Story

_No response_

### Acceptance Criteria

Transform the MockCryptoRng structure into an Rng that output hashes of inputs.

_No response_

SimonRastikian:
Closes #213 - the output is not a Sha256 hash but a chacha output
Partially fixes #107 - for improving the MockCryptoRng
Idea is to improve the MockCryptoRng to be tested and used properly in the Snapshot bench.

### DIFF
--- a/src/test_utils.rs
+++ b/src/test_utils.rs
@@ -9,7 +9,7 @@
 
 use k256::elliptic_curve::PrimeField;
 use k256::AffinePoint;
-use rand_core::{CryptoRng, CryptoRngCore, OsRng, RngCore};
+use rand_core::{CryptoRngCore, OsRng};
 use std::collections::HashMap;
 use std::error::Error;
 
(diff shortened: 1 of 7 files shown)


src/test_utils.rs
```diff
@@ -24,8 +24,35 @@ use crate::{ecdsa, eddsa, ParticipantList};
...
+    }
+    fn fill_bytes(&mut self, dest: &mut [u8]) {
+        self.0.fill_bytes(dest);
+    }
+    fn try_fill_bytes(&mut self, dest: &mut [u8]) -> Result<(), rand::Error> {
+        self.0.try_fill_bytes(dest)
+    }
+}
+
+impl CryptoRng for MockCryptoRng {}
```
gilcu3: Notice that now the implementation is not doing anything, so something as simple as:
```suggestion
pub type MockCryptoRng = ChaCha12Rng;
```
or even using directly `ChaCha12Rng` is enough

PR #218, opened 2025-11-11
Simon/snapshots

### LINKED ISSUE #145: [Benchmark] [Participant Simulation] Protocols Derandomization: Networking Layer Change
As described in #138 , derandomization aims at taking a snapshot of a (any) protocol run.
For this we require changing of the networking layer or the protocol running function to enable writing in an external (ephemeral) file all the data **sent** by the participants.

The snapshot will consist of a set of files, one per party, recording all the messages sent by the corresponding party and in order.
For example: taking the snapshot for participant  Pj  means dynamically creating a file sent_pj which will contain records in the following spirit:
           Pj sent to P0: <bytes>
           Pj sent to P8: <bytes>
           Pj sent to P0: <bytes>                 -------- yes this repetition could be that round 2 started before round 1 ends
           Pj sent to P1: <bytes>
               etc...

To preserve the order, every sent function should take hold of a lock to prevent any concurrent writing in the file.

Acceptance Criteria:
* Create one file per participant
* Append sent data per participant in the file in a precise encoding manner
* Create a lock system for the files
* Implement decoding function for the files that is able to read the files and properly collect the data.
* Test encode/decode

SimonRastikian:
Closes #145

### DIFF
--- a/src/test_utils/mockrng.rs
+++ b/src/test_utils/mockrng.rs
@@ -51,7 +51,7 @@ pub fn create_rngs(
 pub mod test {
     use super::create_rngs;
     use crate::test_utils::{generate_participants, MockCryptoRng};
-    use rand::RngCore;
+    use rand::{RngCore, SeedableRng};
 
     #[test]
     fn test_clone_rngs() {
--- a/src/test_utils/snapshot.rs
+++ b/src/test_utils/snapshot.rs
@@ -107,7 +107,7 @@ mod test {
     use crate::SigningShare;
     use frost_secp256k1::VerifyingKey;
     use k256::ProjectivePoint;
-    use rand_core::CryptoRngCore;
+    use rand_core::{CryptoRngCore, SeedableRng};
 
     fn generate_random_received_snap(rng: &mut impl CryptoRngCore) -> ReceivedMessageSnapshot {
         let from = Participant::from(rng.next_u32());


src/test_utils/run_protocol.rs
```diff
@@ -51,6 +52,69 @@ pub fn run_protocol<T>(
     Ok(out)
 }
 
+/// Like [`run_protocol()`], except that it snapshots all the communication.
+/// Runs a protocol and snapshots all the communication
+pub fn run_protocol_with_snapshots<T>(
+    mut ps: Vec<(Participant, Box<dyn Protocol<Output = T>>)>,
```
gilcu3: could we use a more descriptive name here? that way maybe some comments below would become unnecessary

src/test_utils/run_protocol.rs
gilcu3: I think protocol instead of run_protocol would be a more appropriate name for this file

src/test_utils/snapshot.rs
```diff
@@ -0,0 +1,279 @@
...
+        from: Participant,
+        message: MessageData,
+    ) -> Option<()> {
+        self.snapshots
+            .get_mut(&to)
+            .map(|snapshot| snapshot.push_message(from, message))
+    }
+
+    /// Reads the next message stored in the snapshot of a particular participant given as input
+    pub fn read_next_snapshot(
```
gilcu3:
```suggestion
    pub fn read_next_snapshot_for_participant(
```

src/test_utils/snapshot.rs
```diff
@@ -0,0 +1,279 @@
...
+        let mut psnap = ParticipantSnapshot::new_empty();
+        let mut rvec = Vec::new();
+        for _ in 0..50 {
+            let received_snap = generate_random_received_snap(&mut OsRng);
+            psnap.push_received_message_snapshot(received_snap);
+        }
+        for _ in 0..50 {
+            let (from, message) = psnap.read_next_message().unwrap();
+            let read_message = ReceivedMessageSnapshot::new(from, message);
+            rvec.push(read_message);
```
gilcu3: to make the test more clear, this should be in the above cycle, as in the previous test

src/test_utils/snapshot.rs
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+    fn prepare_keys(p: Participant, f: &Polynomial, big_x: ProjectivePoint) -> KeygenOutput {
+        let private_share = f.eval_at_participant(p).unwrap();
+        let verifying_key = VerifyingKey::new(big_x);
+        KeygenOutput {
+            private_share: SigningShare::new(private_share.0),
+            public_key: verifying_key,
+        }
+    }
```
gilcu3: this function seems out of place, as it is not related to snapshots. Aren't we doing exactly the same thing elsewhere?

src/test_utils/snapshot.rs
```diff
@@ -0,0 +1,279 @@
...
+    fn compare_presign_outputs(o1: &PresignOutput, o2: &PresignOutput) -> bool {
+        o1.alpha == o2.alpha
+            && o1.beta == o2.beta
+            && o1.c == o2.c
+            && o1.e == o2.e
+            && o1.big_r == o2.big_r
+    }
+
+    #[test]
+    fn test_snapshot_on_ecdsa_protocol() {
```
gilcu3: I think this test does not belong to this file, as here we are just implementing general snapshot structs. Not sure what the proper place would be though.

src/test_utils/run_protocol.rs
```diff
@@ -51,6 +52,69 @@ pub fn run_protocol<T>(
...
+                    Action::Return(r) => {
+                        out.push((ps[i].0, r));
+                        false
+                    }
+                }
+            } {}
+        }
+    }
+    Ok((out, protocol_snapshots))
+}
```
gilcu3: This function has an almost identical implementation to the previous one. It would be nice to refactor them so that the logic is only implemented once (in a third function)

src/test_utils/snapshot.rs
```diff
@@ -0,0 +1,279 @@
...
+    message: MessageData,
+}
+
+impl ReceivedMessageSnapshot {
+    fn new(from: Participant, message: MessageData) -> Self {
+        Self { from, message }
+    }
+}
+
+/// Registers a particular participant's view of the received messages
```
netrome: Nit: "Registers" sounds confusing here since this is a data structure, not a behavior.

```suggestion
/// A participant's view of their received messages
```

src/test_utils/snapshot.rs
netrome: Allow me to also add:

```suggestion
    pub fn read_next_message_for_participant(
```

src/test_utils/snapshot.rs
```diff
@@ -0,0 +1,279 @@
...
+        let consumption = rngs.iter_mut().map(RngCore::next_u64).collect::<Vec<_>>();
+        let clone_consumption = clone_rngs
+            .iter_mut()
+            .map(RngCore::next_u64)
+            .collect::<Vec<_>>();
+
+        for (c1, c2) in consumption.iter().zip(clone_consumption.iter()) {
+            assert_eq!(c1, c2);
+        }
+    }
```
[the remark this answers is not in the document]
SimonRastikian: Ok agree.

src/test_utils/snapshot.rs
SimonRastikian: I think it is important to have this test. And as long as we cannot find a better place for it we gonna keep it here.

src/test_utils/snapshot.rs
gilcu3: I did not mean to remove a cycle, but to put this line
```
rvec.push(read_message);
```
in the previous cycle than the one it is now, which seems more natural

SimonRastikian: But I care about testing the read function (twice), not the reception. In fact if I push this line somewhere to the previous cycle, then I should push the received message not the one read. This also imply that the second cycle is (kind of) useless (aside from changing the index).

src/test_utils/snapshot.rs
```diff
@@ -0,0 +1,279 @@
...
+        let from = Participant::from(1u32);
+        let message = b"Hello World".to_vec();
+        let received_snap_1 = ReceivedMessageSnapshot::new(from, message.clone());
+        let from = Participant::from(2u32);
+        let received_snap_2 = ReceivedMessageSnapshot::new(from, message);
+        assert_ne!(received_snap_1, received_snap_2);
+        let message = b"Ciao World".to_vec();
+        let received_snap_3 = ReceivedMessageSnapshot::new(from, message);
+        assert_ne!(received_snap_2, received_snap_3);
+    }
```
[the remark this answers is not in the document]
gilcu3: arent you testing equality and inequality of `ReceivedMessageSnapshot` objects ? It does not matter how you build them, equality is guaranteed by the derived methods, so not worth having extra tests for it IMO

src/test_utils/snapshot.rs
```diff
@@ -0,0 +1,279 @@
...
+        }
+    }
+
+    fn compare_presign_outputs(o1: &PresignOutput, o2: &PresignOutput) -> bool {
+        o1.alpha == o2.alpha
+            && o1.beta == o2.beta
+            && o1.c == o2.c
+            && o1.e == o2.e
+            && o1.big_r == o2.big_r
+    }
```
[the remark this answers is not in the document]
gilcu3: I think those two traits are fine, they should not change any behavior anywhere else, so please add them instead

src/test_utils/mockrng.rs
```diff
@@ -26,3 +27,41 @@ impl RngCore for MockCryptoRng {
...
+    pub fn create_rngs(
+        participants: &[Participant],
+        seed: &mut impl CryptoRngCore,
+    ) -> Vec<MockCryptoRng> {
+        let rngs = participants
+            .iter()
+            .map(|_| MockCryptoRng::seed_from_u64(seed.next_u64()))
+            .collect::<Vec<_>>();
+        rngs
+    }
```
gilcu3: Now that I think about this, it makes sense to probably expose this function, not putting it inside tests, as this is the test_utils module after all

gilcu3 (approved):
Thank you for the updates! Left a couple of nits

src/test_utils/snapshot.rs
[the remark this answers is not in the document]
SimonRastikian: Solved

gilcu3 (approved):

PR #220, opened 2025-11-12
feat: added uniform threshold lower bound 2 in keygen and reshare + tests

### LINKED ISSUE #160: Resharing with a new participant and threshold equal to one will always fail
This is a corner case that is not used in our systems but it's is good to have it patched anyway. The problem being that the reshare protocol will always fail when the threshold is one and a new participant joins the protocol. 
When a new participant is participating in the reshare protocol, it sets its secret to zero, and creates a polynomial of degree threshold-1. However, since the `Polynomial::new` function returns an error for the constant zero polynomial, this resharing protocol will always fail.

This exists in `do_keyshare`: 
```
let secret_coefficients = 
        Polynomial::<C>::generate_polynomial(Some(secret), threshold - 1, rng)?; 
```

A new participant joins the MPC nodes, where the threshold is set to one. However, the new participant can never generate the secret polynomial, and the protocol will always fail. 

Acceptance criteria:
* Add validation to the threshold value, as well as checked arithmetic to operations on it. Add edge case tests exercising cases where the operation should fail, and where it should still succeed. 

* Link to #18: formally specify the resharing and refresh protocols, ensuring that all necessary checks are described. Then, annotate the implementation with the specification steps, ensuring that no step is missing from the implementation.

gilcu3:
Closes #160

- most new code is just tests for the limit cases for keygen and reshare

### DIFF
--- a/src/dkg.rs
+++ b/src/dkg.rs
@@ -706,7 +706,8 @@ pub mod test {
         assert_eq!(<C::Group as Group>::generator() * x, pub_key);
     }
 
-    pub fn test_keygen_threshold_limits<C: Ciphersuite>()
+    #[allow(non_snake_case)]
+    pub fn keygen__should_fail_if_threshold_is_below_limit<C: Ciphersuite>()
     where
         <C::Group as Group>::Element: std::fmt::Debug + std::marker::Send,
         <<C::Group as Group>::Field as Field>::Scalar: std::marker::Send,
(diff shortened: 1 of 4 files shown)


src/dkg.rs
```diff
@@ -763,4 +789,39 @@ pub mod test {
         }
         assert_eq!(<C::Group as Group>::generator() * x, pub_key.to_element());
     }
+
+    pub fn test_reshare_threshold_limits<C: Ciphersuite>()
```
netrome: Nice with the generic tests!

Nit: Pushing for the naming convention we haven't formalized yet

```suggestion
    pub fn reshare__should_fail_if_threshold_is_below_limit<C: Ciphersuite>()
```

netrome (approved):

PR #225, opened 2025-11-17
chore: added deterministic tests

### LINKED ISSUE #107: Deterministic unit tests

gilcu3:
Closes #107 

- This removes the usage of non-determinism from the code-base, except from the integration tests. I believe using `OsRng` in these files is preferred, as currently they are the closest we have to usage examples.
- It checks that the results of the main functions in the crate are deterministic, which covers all functions (albeit indirectly). I think this is good enough atm.
- This PR is meant to be merged after the test_utils refactor #217. It should not have any conflict with the snapshots #218
- This PR had no choice but slightly breaking the API of test_utils, as some functions were using OsRng internally, and now expect the RNG as a parameter.


src/ecdsa/ot_based_ecdsa/triples/generation.rs
gilcu3: this is the only file where the changes touch production code. Without it the code was not testable, as the only RNG that implements Copy was OsRng. In any case, I believe the current implementation is only correct when using OsRng, so I plan to address this in a follow up, that would require an unrelated refactoring here, and the removal of all `Clone` requirements in the traits

benches/inversion.rs
```diff
@@ -2,18 +2,19 @@
...
 use frost_secp256k1::{Secp256K1ScalarField, Secp256K1Sha256};
-use rand_core::OsRng;
+use rand::SeedableRng;
 use std::hint::black_box;
-use threshold_signatures::batch_invert;
+use threshold_signatures::{batch_invert, test_utils::MockCryptoRng};
 
 fn bench_inversion(c: &mut Criterion) {
     let mut group = c.benchmark_group("Single_vs_Batch_Inversion");
+    let mut rng = MockCryptoRng::seed_from_u64(42);
```
DSharifi: Small nit:
Does it make sense to have a shared seed and define it as a const?

src/confidential_key_derivation/test.rs
```diff
@@ -1,25 +1,37 @@
...
-    crate::dkg::test::test_keygen::<C>(&participants, threshold);
+    let result = crate::dkg::test::test_keygen::<C, _>(&participants, threshold, &mut rng);
+    let serialized_result = serde_json::to_string(&result).unwrap();
+    assert_eq!(serialized_result, "[[0,{\"private_share\":\"a8d9bc9f09e6085c448851e60a980981ceac312b0bf801a510a13eb44c7b5d0e\",\"public_key\":\"acede0416bc326a4a52a52f74391b1bbd131a45950bfe4ee7be1cfb18bb0cb8e2abba6fad27658c2c4186244a379e16108f1db1e9168d8763e16774a427e706fa5a51703eb84ff27776758270fcb0dae883c6e5bacbe9a639064c2fd466d8357\"}],[1,{\"private_share\":\"560c12b2bbdb00df26df931883e5f8731a6ed7bda7a7d4530c40712a7fa88c28\",\"public_key\":\"acede0416bc326a4a52a52f74391b1bbd131a45950bfe4ee7be1cfb18bb0cb8e2abba6fad27658c2c4186244a379e16108f1db1e9168d8763e16774a427e706fa5a51703eb84ff27776758270fcb0dae883c6e5bacbe9a639064c2fd466d8357\"}],[2,{\"private_share\":\"043f67c46dd1f8610936d64afb32e866662f7d504457a70208dfa3a0b1d5bb42\",\"public_key\":\"acede0416bc326a4a52a52f74391b1bbd131a45950bfe4ee7be1cfb18bb0cb8e2abba6fad27658c2c4186244a379e16108f1db1e9168d8763e16774a427e706fa5a51703eb84ff27776758270fcb0dae883c6e5bacbe9a639064c2fd466d8357\"}]]");
```
DSharifi: Why are we introducing json serialization checks to these tests? If we want serialization tests, does it not make sense to have that as separate tests?

src/ecdsa/ot_based_ecdsa/triples/batch_random_ot.rs
```diff
@@ -406,25 +409,28 @@ mod test {
         let r = Participant::from(1u32);
         let comms_s = Comms::new();
         let comms_r = Comms::new();
-
+        let rng_1 = R::seed_from_u64(rng.next_u64());
```
DSharifi: Nice.

alt text

DSharifi (dismissed):
Thanks a lot for these changes! :)

src/confidential_key_derivation/protocol.rs
```diff
@@ -220,7 +218,8 @@ mod test {
 
         let mut private_shares = Vec::new();
         for p in &participants {
-            let private_share = SigningShare::new(Scalar::random(&mut rng));
+            let mut rng_p = MockCryptoRng::seed_from_u64(rng.next_u64());
```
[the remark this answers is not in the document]
gilcu3: oh, it was simpler, p is just the variable we use for indexing two lines above. What name would you suggest?

src/confidential_key_derivation/test.rs
netrome: Yeah this feels overkill. While I appreciate the idea of testing the determinism, I don't think tests necessarily need to assert or prove that they are deterministic. It just should be the standard way we implement these tests.

Though it does open up the question if we can add some type of automation to ensure tests stay deterministic. I've never had that in the past, but from a pure idealistic standpoint it would be nice if we could have some way to prevent people from accidentally introducing non-determinism in tests.

DSharifi (approved):

barakeinav1:
I did a brief review (loved your use of 42) . but I don't feel I have any useful feedback here.
I can approve as a rubber stamp, or you can wait for another reviewer. 
let me know.

barakeinav1 (dismissed):

SimonRastikian (dismissed):
Should be good overall,
Couple of comments about rng traits. Also I am not a big fan of the added files...

benches/naive_ot_based_ecdsa.rs
```diff
@@ -111,13 +114,17 @@ type PreparedTriples = Vec<(
     Participant,
     Box<dyn Protocol<Output = Vec<(TripleShare, TriplePub)>>>,
 )>;
-fn prepare_triples(participant_num: usize) -> PreparedTriples {
+fn prepare_triples<R: CryptoRngCore + SeedableRng + Send + 'static>(
+    participant_num: usize,
+    rng: &mut R,
```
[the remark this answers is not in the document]
SimonRastikian: rng: impl CryptoRngCore + SeedableRng + Send + 'static

netrome (dismissed):

PR #236, opened 2025-11-25
fix: get CI unstuck

### LINKED ISSUE #235: Get CI unstuck
### Background

Currently, the PR in #218 cannot be merged because it fails CI, while the code works fine locally. The issue seems to be related to the usage of the old version of ubuntu in CI.

### User Story

_No response_

### Acceptance Criteria

Update the CI so that the PR can be merged.

[thread](https://nearone.slack.com/archives/C0912BTG51T/p1764009424162399)

gilcu3:
Closes #235

### DIFF
--- a/tests/eddsa.rs
+++ b/tests/eddsa.rs
@@ -45,8 +45,8 @@ fn run_sign(
 
 #[test]
 fn test_sign() {
-    let participants = generate_participants(11);
-    let max_malicious = 5;
+    let participants = generate_participants(5);
+    let max_malicious = 3;
     let threshold = max_malicious + 1;
     let keys = run_keygen::<C>(&participants, threshold);
     assert_eq!(keys.len(), participants.len());
--- a/src/eddsa/sign.rs
+++ b/src/eddsa/sign.rs
@@ -449,7 +449,7 @@ mod test {
     #[test]
     fn test_reshare_sign_more_participants() {
         let mut rng = MockCryptoRng::seed_from_u64(42);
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -58,24 +59,24 @@ jobs:
       - name: Run cargo-nextest
         run: cargo nextest run --locked --all-features --cargo-profile test-release
 
-      - name: Run cargo test debug
-        run: cargo test --locked --all-features --all-targets
-
   ci-coverage:
     name: Coverage
-    runs-on: warp-ubuntu-2204-x64-8x
+    runs-on: warp-ubuntu-2404-x64-8x
```
gilcu3: this was provoking CI failures in #218

tests/eddsa.rs
```diff
@@ -45,8 +45,8 @@ fn run_sign(
 
 #[test]
 fn test_sign() {
-    let participants = generate_participants(11);
-    let max_malicious = 5;
+    let participants = generate_participants(5);
+    let max_malicious = 3;
```
gilcu3: the previous parameters were making the coverage test quite slow for no good reason

kevindeforth (approved):

tests/eddsa.rs
gilcu3: yeah I think that would be more suitable for benchmarks. The reason why it becomes so slow only in coverage tests is something worth figuring out

Issue #136, opened 2025-10-03
[Benchmark] Participant Simulation: Generic Description


SimonRastikian:
### Background

There exist many strategies that help calculate properly the amount of time needed for the protocol to execute.
We decided that it is best to do so by simulating all the participants except for one single participant. This allows syncing between the different schemes based on the maximal number of malicious parties (for more info see #135 ).

### User Story

Without loss of generality, we assume that the participant to benchmark is P0
The simulation should be divided into two steps.
- In the first step, an rng is fixed and all the participants are ran honestly as in a normal setup. The first step should record all the messages sent to P0 and store them in a database **DB** keeping track of the communications between each participant and P0.
- In the second step, would run all over again the protocol, however, this time, only between a simulation and P0. The simulation function would only listen to P0 and upon receiving a message, sends back the next item in **DB**. It is essential to run P0 using the exact same randomness used in step 1.

Note that: The benchmarking would only care for how much time P0 in Step 2 takes to finalize.

PS: 
* Latency could be modeled using a wait function before sending and after receiving. 
* The amount of data sent is equals to the sum of items in the database
* The end-to-end protocol time should be calculated with the simulate-function (which could allow for latency using wait function calls)

### Acceptance Criteria

* Derandomize the participants in step 1.
* Store the data sent to P0 in a database.
* Implement  `simulated_participants` with input DB.  This function listens to P0 and sends in the items stored in DB in order and following participants, i.e., simulate a single entity that acts as if it was many different participants.
* The benchmark should include the call of simulated_participant function.

### Resources & Additional Notes

References:
#135 
#134

Issue #134, opened 2025-10-03
[Benchmark] [OT-ECDSA] Generic OT-Based ECDSA Benchmark Issue


SimonRastikian:
### Background

We would like to benchmark the OT-Based ECDSA as we have little objective measurement tools.

### User Story

The details of how to benchmark should be described in multiple issues explaining:
1) How to simulate participants independently in the three different phases (Triples, Presign, Sign)
2) How to measure the time taken by a single (real)  participant in the three different phases (Triples, Presign, Sign)
3) What to change in the code to be able to add latency
4) How to measure the computation time per participant
5) How to measure the amount of data transmitted
6) How to measure the size of storage in between phases   

### Acceptance Criteria

The benchmark should cover Triple Generation, Presigning and Signing phases.
The number of participants should be the max number of malicious parties + 1

The measured parameters are:
* End-to-end computation time (with/without latency) per participant
* Total amount of data sent per participant
* Computation time per participant (without send and receive)
* Size of storage

The technique used to measure these factors is by participants simulation (described in #135 )

### Resources & Additional Notes

Even though the scheme allows for robustness, changing the number of parties between the phases is not included in the scope of this issue.

Reference: "Mother of all Benchmark Issues" #135

0xsecaas:
> The number of participants should be the max number of malicious parties + 1

so `N=t` or `N = f +1`?

SimonRastikian:
@0xsecaas could I am not sure about your notation but the idea is that we will fix the same accepted number of malicious parties for both ECDSA scheme.
We will run each scheme with the minimal number of parties such that the signature generated is valid

PR #240, opened 2025-12-01
Benchmarks with latency

SimonRastikian:
Allows modifying the benchmark parameters Iterations and Latency.

Computes the schemes with modifiable latency in milliseconds.
Contributes to the latency section of #140 #134  and naturally to #136

### DIFF
--- a/benches/bench_utils.rs
+++ b/benches/bench_utils.rs
@@ -29,7 +29,7 @@ pub static MAX_MALICIOUS: LazyLock<usize> = std::sync::LazyLock::new(|| {
         .unwrap_or(6)
 });
 
-// fix malicious number of participants
+// fix number of samples
 pub static SAMPLE_SIZE: LazyLock<usize> = std::sync::LazyLock::new(|| {
     env::var("SAMPLE_SIZE")
         .ok()


netrome (changes requested):
I don't understand how the current sleep call helps us simulate network latency. This looks like the wrong place to sleep in my view. I'd like to get more clarification on that when you have the time.

benches/naive_robust_ecdsa.rs
```diff
@@ -31,16 +35,17 @@ fn bench_presign(c: &mut Criterion) {
...
-    group.measurement_time(std::time::Duration::from_secs(300));
+    group.sample_size(*SAMPLE_SIZE);
 
     let preps = robust_ecdsa_prepare_presign(num, &mut rng);
     let result = run_protocol(preps.protocols).expect("Prepare sign should not");
     let pk = preps.key_packages[0].1.public_key;
 
     group.bench_function(
-        format!("robust_ecdsa_sign_naive_MAX_MALICIOUS_{max_malicious}_PARTICIPANTS_{num}"),
+        format!("robust_ecdsa_sign_naive_MAX_MALICIOUS_{max_malicious}_PARTICIPANTS_{num}_LATENCY_{latency}"),
```
gilcu3: Before accepting or refusing this change I need to know what was the idea to include it. Before we had a constant time duration for the benchmark, and now we are letting the users set the sample size. As the parameter input method we use are not great (we are forced to use env variables by criterion), then I would prefer to remain as before (const total time) or move to a constant number of samples, or use criterion's default, instead of making the implementation more verbose. 

If we have a good reason to add sample size let me know.

gilcu3 (changes requested):
I have trouble understanding what we achieve with the PR. 

- It is not clear why we need a sample size parameter, instead of using criterion's default or a fixed amount of time as before. 

- It also looks like the benchmark latency measurement is not a measurement, but an estimation, in which case I think it should be in the docs but not included in the criterion benchmarks.

benches/naive_robust_ecdsa.rs
gilcu3: I see, I thought the previous `group.measurement_time(std::time::Duration::from_secs(300));` referred to maximum allowed total time. From criterion docs:
```
pub fn measurement_time(&mut self, dur: Duration) -> &mut Self

Criterion will attempt to spent approximately this amount of time measuring each benchmark on a best-effort 
basis. If it is not possible to perform the measurement in the requested time (eg. because each iteration of the 
benchmark is long) then Criterion will spend as long as is needed to collect the desired number of samples. With 
a longer time, the measurement will become more resilient to interference from other programs.
```

so why instead we put a constant, say 15 as you mention, for the benchmarks that we know otherwise would take too long with the default value, and leave the rest as is? I cannot think of a user that would want a different amount of samples, and worst case they could simply fork, modify and run.

gilcu3: as we discussed having this parameter might be useful for the slower benchmarks, as is explained in the current README. marking this as solved

src/test_utils/protocol.rs
```diff
@@ -193,5 +198,6 @@ pub fn run_simulated_protocol<T>(
             out = Some(output);
         }
     }
+    std::thread::sleep(duration);
     out.ok_or_else(|| ProtocolError::Other("out is None".to_string()))
 }
```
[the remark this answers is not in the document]
SimonRastikian: Done!

gilcu3 (dismissed):
Thank you!

two leftovers still need to be cleaned, but that should be quick so I am approving already

gilcu3 (dismissed):

netrome (approved):

benches/bench_utils.rs
```diff
@@ -29,6 +29,14 @@ pub static MAX_MALICIOUS: LazyLock<usize> = std::sync::LazyLock::new(|| {
         .unwrap_or(6)
 });
 
+// fix malicious number of participants
```
gilcu3: just noticed this comment is not accurate

PR #226, opened 2025-11-18
Simon/simulation

SimonRastikian:
Closes Parts of #138  #134 #137 #139

None of these (more generic) issues are fully closed except for #137  as the others include adding latency, measuring the quantity of data sent, documenting the bench model, and adding EdDSA into the benchmarks.
In a nutshell, this PR introduces the simulator and the protocol simulation functionalities, then performs the advanced benchmarking of ECDSA scheme.

PR #242, opened 2025-12-03
fix: simulation rngs fix

gilcu3:
Follow up to #226


gilcu3:
> Forgive me but I do not get what you are trying to solve...

This just shows that you dont need to compute the rngs before calling `ot_ecdsa_prepare_triples`. This was the only place in the code were it was actually needed, but after the modification I propose here it is no longer the case.

gilcu3:
closing this, as it was only a PoC. Feel free to use the same idea/code on the actual branch.

PR #254, opened 2025-12-10
Simon/documenting dkg

### LINKED ISSUE #99: Add doc for dkg
### Background

Currently we are using the same DKG protocol for all schemes, but it is not documented which one it is.

### User Story

As a user I would like to understand how the library works, at least on a high level, without having to read the code.

### Acceptance Criteria

A new doc is added explaining how the DKG is done.

_No response_

SimonRastikian:
Closes #99


docs/dkg.md
```diff
@@ -0,0 +1,139 @@
...
+a stronger notion of security than the one promised by plain PedPop.
+
+PedPop+ is a five rounds protocol and makes use in three of its rounds of a reliable broadcast channel. A reliable broadcast is a 3 round protocol,
+implying that the total number of PedPop+ rounds 11. The broadcast channel is implemented in `src/protocol/echo_broadcast.rs`.
+
+The implemented DKG serves as a generic one that can be used with multiple different underlying elliptic curves. We thus use it with `Secp256k1` for ECDSA schemes, `Curve25519` for EdDSA scheme, and `BLS12_381` for the confidential key derivation functionality.
+
+## Keygen, Reshare and Refresh
+
+The core of the dkg protocol is implemented in a subfunction called `do_keyshare` and serves for three applications:
```
netrome: What do you mean by "subfunction"? Should we just call it "function" instead perhaps?

docs/dkg.md
```diff
@@ -0,0 +1,139 @@
...
+
+* Key resharing: denoted in the implementation with `reshare`. It allows for a set of participants who already own valid private shares to kick away other participants from the pool, create fresh shares for new participants i.e. new joiners to the pool, and/or change the **cryptographic threshold** described in section [Types of Thresholds](#types-of-thresholds).
+
+* Key refresh: denoted in the implementation with `refresh`. It is a special case of the key resharing in which the set of participants stays the same before and after the protocol run and with no changes to the crypto. The goal being that each participant would refresh their signing share without modifying the master public key.
+
+## Types of Thresholds
+
+There are two types of thresholds one has to be aware of: the **asynchronous distributed systems threshold** a.k.a. the **BFT threshold**, and the **cryptography threshold** a.k.a. the **reconstruction threshold**.
+
+The BFT threshold states that the maximum number of faulty nodes a distributed system ($\mathsf{MaxFaulty}$) can tolerate while still reaching consensus is at most one-third of the total number of participants $N$. More specifically:
```
netrome: Hmm, I like that we're getting into the threshold types here but I wonder if it's relevant for the DKG algorithm. `MaxFaulty` is unused so I think we could move this part of the discussion to an appendix or similar since it's still interesting. For the purpose of understanding the DKG we only need to know the `MaxMalicious` parameter which in this case is the reconstruction lower bound - 1 as per our conversation yesterday.

Also, I think we take the reconstruction bound as input in the code but we talk abut max malicious here. Is that correct? I wonder if we should consider migrating to use max malicious as input to the dkg protocol in the future (which is a bit of a tricky change to do safely).

docs/dkg.md
```diff
@@ -0,0 +1,139 @@
...
+
+$$\mathsf{MaxFaulty} \leq \frac{N - 1}{3}$$
+
+The cryptography threshold refers to the maximum number of necessay malicious parties ($\mathsf{MaxMalicious}$) a certain scheme can handle without compromising on the security and assuming the existance of an underlying reliable broadcast channel. $\mathsf{MaxMalicious}$ is scheme dependent and can have a different value than $\mathsf{MaxFaulty}$. For instance, in the OT based ECDSA, $\mathsf{MaxMalicious}$ can be up to $N-1$, but in Robust ECDSA scheme $\mathsf{MaxMalicious}$ must not exceed $\frac{N - 1}{3}$.
```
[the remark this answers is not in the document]
SimonRastikian: Absolutely, oups this slipped

docs/dkg.md
```diff
@@ -0,0 +1,139 @@
+# Distribute Key Generation
```
gilcu3:
```suggestion
# Distributed Key Generation
```

docs/dkg.md
```diff
@@ -0,0 +1,135 @@
+# Distributed Key Generation
+
+We define a variant of the two-round DKG protocol PedPop \[[KG](https://eprint.iacr.org/2020/852.pdf)\].
+Our variant, PedPop+ is less efficient, but achieves a notion of simulatability with aborts,
+a stronger notion of security than the one promised by plain PedPop.
+
+PedPop+ is a five rounds protocol and makes use in three of its rounds of a reliable broadcast channel. A reliable broadcast is a 3 round protocol,
+implying that the total number of PedPop+ rounds 11. The broadcast channel is implemented in `src/protocol/echo_broadcast.rs`.
+
+The implemented DKG serves as a generic one that can be used with multiple different underlying elliptic curves. We thus use it with `Secp256k1` for ECDSA schemes, `Curve25519` for EdDSA scheme, and `BLS12_381` for the confidential key derivation functionality.
```
gilcu3: the usual spelling for bls12-381 curve is with hyphen
```suggestion
The implemented DKG serves as a generic one that can be used with multiple different underlying elliptic curves. We thus use it with `Secp256k1` for ECDSA schemes, `Curve25519` for EdDSA scheme, and `BLS12-381` for the confidential key derivation functionality.
```

docs/dkg.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Keygen, Reshare and Refresh
+
+The core of the dkg protocol is implemented in a function called `do_keyshare` and serves for three applications:
```
gilcu3: DKG was upper case before
```suggestion
The core of the DKG protocol is implemented in a function called `do_keyshare` and serves for three applications:
```

docs/dkg.md
```diff
@@ -0,0 +1,135 @@
...
+
+* Key refresh: denoted in the implementation with `refresh`. It is a special case of the key resharing in which the set of participants stays the same before and after the protocol run and with no changes to the crypto. The goal being that each participant would refresh their signing share without modifying the master public key.
+
+## Types of Thresholds
+
+There are two types of thresholds one has to be aware of: the **asynchronous distributed systems threshold** a.k.a. the **BFT threshold**, and the **cryptography threshold** a.k.a. the **reconstruction threshold**.
+
+The BFT threshold states that the maximum number of faulty nodes a distributed system ($\mathsf{MaxFaulty}$) can tolerate while still reaching consensus is at most one-third of the total number of participants $N$. More specifically:
+
+$$\mathsf{MaxFaulty} \leq \frac{N - 1}{3}$$
```
gilcu3: the BFT threshold is not mentioned anywhere else, so it seems that this could be extra information for an appendix and not really necessary at the beginning of the doc, as it might confuse the reader instead of providing information. In the appendix we could mention that any threshold we use must be above $\frac{N - 1}{3}$ because of the underlying broadcast channel implementation requirements

docs/dkg.md
```diff
@@ -0,0 +1,135 @@
...
+### DKG and thresholds
+
+Due to the fact that PedPop+ utilizes reliable broadcast channel to securely generate private shares, it thus lies on the edge between the asynchronous distributed systems and cryptography. For this reason, we set
+$\mathsf{MaxFaulty} = \frac{N - 1}{3}$ as an invariable parameter and allow our key generation and key resharing protocols to fix/modify only the $\mathsf{MaxMalicious}$ threshold depending on the scheme requirements and on the library user's choice.
+
+## Technical Details: Key Generation & Key Resharing
+
+Let $P_1, \cdots P_N$ be $N$ different participants, and $\mathsf{MaxMalicious}$ be the desired cryptography threshold. Let $H_1, H_2, H_3$ be domain separated hash functions.
+
+We define PedPop+ key generation as follows; All the instruction preceeded with `+++` are added to the key generation, transforming it to the key resharing protocol.
```
gilcu3: in English you do not use capitalization after semi-colons.
```suggestion
We define PedPop+ key generation as follows, where all the instructions preceded with `+++` are added to the key generation, transforming it to the key resharing protocol.
```

docs/dkg.md
```diff
@@ -0,0 +1,135 @@
...
+2. `+++` The old participants set $\mathit{OldSigners}$ that held valid private shares prior to the key resharing.
+
+3. `+++` The old master public key $\mathit{OldPK}$ that the $\mathit{OldSigners}$ held prior to the key resharing.
+
+4. `+++` The old cryptography threshold $\mathsf{OldMaxMalicious}$ prior to the key resharing.
+``
+
+### Round 1
+
+1.1 Each $P_i$ asserts that $1 < \mathsf{MaxMalicious} < N$.
```
gilcu3: the implementation does this in `assert_keygen_invariants`, but uses `threhold`, so here we should use the same term

gilcu3 (changes requested):
This is a partial review, but already finds some blockers, so I guess we can work on this again after the holidays

- The docs refer to `MaxMalicious` instead of `threshold` that is used all over the implementation. This mismatch defeats the purpose of the doc, which is mainly describing the implemented protocol
- The current doc merges key generation and resharing. This makes following the doc much harder, and requires extra care. For example, the functions to assert keygen invariants and the one for asserting resharing invariants are not the same: one is `assert_keygen_invariants` and the other `reshare_assertions`, but the doc describes both as the same. I think these distinctions need to be explicit in the doc. Thinking about this in more generality, I think our current implementation is overly complex precisely because these two functionalities are implemented by the same function using `Option` to handle the differences in many places. Long term we should evaluate if that is better than simply having each functionality implemented in its own function.

docs/dkg.md
SimonRastikian: This is my french reflex :fr:

docs/dkg.md
gilcu3: I am a bit torn about adding docs that are not in sync with implementation. If we ultimately decide to use `MaxMalicious` everywhere, then we could fix both at the same time. But until then, I think we should use the old name `threshold` as long as it does not contradict anything.

docs/dkg.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Types of Thresholds
+
+There are two types of thresholds one has to be aware of: the **asynchronous distributed systems threshold** a.k.a. the **BFT threshold**, and the **cryptography threshold** a.k.a. the **reconstruction threshold**.
+
+The BFT threshold states that the maximum number of faulty nodes a distributed system ($\mathsf{MaxFaulty}$) can tolerate while still reaching consensus is at most one-third of the total number of participants $N$. More specifically:
+
+$$\mathsf{MaxFaulty} \leq \frac{N - 1}{3}$$
+
+The cryptographic threshold refers to the maximum number of malicious parties ($\mathsf{threshold}$) that a scheme can tolerate without compromising security, assuming the existence of an underlying reliable broadcast channel. $\mathsf{threshold}$ is scheme dependent and can have a different value than $\mathsf{MaxFaulty}$. For instance, in the OT based ECDSA, $\mathsf{threshold}$ can be up to $N-1$, but in Robust ECDSA scheme $\mathsf{threshold}$ must not exceed $\frac{N - 1}{2}$.
```
gilcu3: in the implementation afaik we use $\mathsf{threshold} = \mathsf{MaxFaulty} + 1$. There same must be used here

gilcu3 (changes requested):
Requesting changes because there is still (at least) one inconsistency in the use of `threshold`. Please verify there are no leftovers from previous iterations.

> The current doc merges key generation and resharing. This makes following the doc much harder, and requires extra care. For example, the functions to assert keygen invariants and the one for asserting resharing invariants are not the same: one is assert_keygen_invariants and the other reshare_assertions, but the doc describes both as the same. I think these distinctions need to be explicit in the doc. Thinking about this in more generality, I think our current implementation is overly complex precisely because these two functionalities are implemented by the same function using Option to handle the differences in many places. Long term we should evaluate if that is better than simply having each functionality implemented in its own function.

That was part of the last review, but I could not find any response. WDYT?

docs/dkg.md
SimonRastikian: You are right! Revisited in [bde965d](https://github.com/near/threshold-signatures/pull/254/commits/bde965d331e907e7fdad5d25b08dd05153afe59a)

docs/dkg.md
```diff
@@ -27,12 +27,12 @@ The BFT threshold states that the maximum number of faulty nodes a distributed s
...
 
 ### DKG and thresholds
 
 Due to the fact that PedPop+ utilizes reliable broadcast channel to securely generate private shares, it thus lies on the edge between the asynchronous distributed systems and cryptography. For this reason, we set
-$\mathsf{MaxFaulty} = \frac{N - 1}{3}$ as an invariable parameter and allow our key generation and key resharing protocols to fix/modify only the $\mathsf{threshold}$ threshold depending on the scheme requirements and on the library user's choice.
+$\mathsf{MaxFaulty} = \frac{N - 1}{3}$ as an invariable parameter and allow our key generation and key resharing protocols to fix/modify only the $\mathsf{threshold}$ dependinthresholdg on the scheme requirements and on the library user's choice.
```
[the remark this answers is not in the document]
SimonRastikian: done

docs/dkg.md
```diff
@@ -0,0 +1,135 @@
+# Distributed Key Generation
+
+We define a variant of the two-round DKG protocol PedPop \[[KG](https://eprint.iacr.org/2020/852.pdf)\].
```
DSharifi: Shold we define DKG before using the acronym in these docs?

docs/dkg.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Round 3
+
+3.1 Each $P_i$ waits to receive $h_i$ from every participant $P_j$.
+
+3.2 Each $P_i$ reliably broadcasts $(C_i, R_i, s_i)$.
+
+### Round 4
+
+4.1 Each $P_i$ waits to receive $(C_j, \pi_j)$ from every participant $P_j$.
```
DSharifi: typo?


```suggestion
4.1 Each $P_i$ waits to receive $(C_j, \p_j)$ from every participant $P_j$.
```

docs/dkg.md
```diff
@@ -0,0 +1,135 @@
...
+
+2.7 Each $P_i$ computes the proof $s_i \gets k_i + f_i(0) \cdot c_i$
+
+2.8 Each $P_i$ generates a hash $h_i \gets H_2(i, C_i, \mathit{sid})$
+
+2.9 Each $P_i$ sends $h_i$ to every participant
+
+### Round 3
+
+3.1 Each $P_i$ waits to receive $h_i$ from every participant $P_j$.
```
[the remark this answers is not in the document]
SimonRastikian: Nice catch

docs/dkg.md
```diff
@@ -0,0 +1,135 @@
...
+
+3.2 Each $P_i$ reliably broadcasts $(C_i, R_i, s_i)$.
+
+### Round 4
+
+4.1 Each $P_i$ waits to receive $(C_j, \pi_j)$ from every participant $P_j$.
+
+4.2 Each $P_i$ computes: $\forall j\in\set{1, \cdots N}, \quad c_j \gets H_3(\mathit{sid}, j, C_j(0), R_j)$.
+
+4.3 Each $P_i$ asserts that: $\forall j\in\set{1, \cdots N}, \quad R_j = s_i \cdot G - c_j \cdot C_j(0)$.
```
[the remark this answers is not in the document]
SimonRastikian: Correct

DSharifi (dismissed):

gilcu3 (dismissed):

netrome (approved):
Thanks for adding the definition of the lagrange coefficient! :pray:

gilcu3 (approved):

PR #264, opened 2025-12-22
fix: reduce index slicing usage, removed from polynomial and participants

### LINKED ISSUE #263: remove index-slicing exceptions from participants and polynomials
### Background

We have index-slicing exceptions in several places where they can be easily removed.

### User Story

_No response_

### Acceptance Criteria

index-slicing exceptions from participants and polynomials are removed. 

_No response_

gilcu3:
Closes #263

### DIFF
--- a/src/crypto/polynomials.rs
+++ b/src/crypto/polynomials.rs
@@ -91,12 +91,12 @@ impl<C: Ciphersuite> Polynomial<C> {
 
     /// Returns the constant term or error in case the polynomial is empty
     pub fn eval_at_zero(&self) -> Result<SerializableScalar<C>, ProtocolError> {
-        Ok(SerializableScalar(
-            self.coefficients
-                .first()
-                .copied()
-                .ok_or(ProtocolError::EmptyOrZeroCoefficients)?,
-        ))
+        let result = self
+            .coefficients
+            .first()
+            .copied()
+            .ok_or(ProtocolError::EmptyOrZeroCoefficients)?;
+        Ok(SerializableScalar(result))
     }
 
     /// Evaluates a polynomial at a certain scalar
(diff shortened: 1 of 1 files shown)


src/crypto/polynomials.rs
```diff
@@ -88,10 +91,12 @@ impl<C: Ciphersuite> Polynomial<C> {
...
-        if self.coefficients.is_empty() {
-            return Err(ProtocolError::EmptyOrZeroCoefficients);
-        }
-        Ok(SerializableScalar(self.coefficients[0]))
+        let result = self
+            .coefficients
+            .first()
+            .copied()
+            .ok_or(ProtocolError::EmptyOrZeroCoefficients)?;
+        Ok(SerializableScalar(result))
```
DSharifi:
```suggestion
        self
            .coefficients
            .first()
            .copied()
            .ok_or(ProtocolError::EmptyOrZeroCoefficients)
```

src/crypto/polynomials.rs
```diff
@@ -502,8 +521,16 @@ where
...
+
+        let coeffs = (0..n)
+            .map(|i| {
+                if i == kronecker_index_value {
+                    SerializableScalar(<C::Group as Group>::Field::one())
+                } else {
+                    SerializableScalar(<C::Group as Group>::Field::zero())
+                }
+            })
+            .collect();
```
DSharifi: How big is `n` usually? 

Building the coefficients instead in an iterator will be a performance hit if this code is part of a hot path;

Can we keep the existing vector allocation and a `get_mut` call instead?

src/crypto/polynomials.rs
gilcu3: We still need the `SerializableScalar` wrap, but if you think putting everything in one line is better, we can do that

src/crypto/polynomials.rs
gilcu3: n is the size of the polynomial, so as bug as the number of participants. Will change as you mention, although I am not 100% sure how much of a performance hit that is.

src/crypto/polynomials.rs
DSharifi: Ahh right, so there's a `map`. Yeah, current approach is fine

src/crypto/polynomials.rs
```diff
@@ -219,22 +228,29 @@ impl<C: Ciphersuite> PolynomialCommitment<C> {
...
+
         // get the number of non-identity coeffs
         let last_non_id = coefcommitments.len() - count;
-        Ok(Self {
-            coefficients: coefcommitments[..last_non_id].to_vec(),
-        })
+
+        let new_coefficients = coefcommitments
+            .get(..last_non_id)
+            .ok_or(ProtocolError::EmptyOrZeroCoefficients)?
```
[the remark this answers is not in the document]
DSharifi: > also notice that last_non_id > coefcommitments.len() is impossible

Yes.

But if the code relies on this invariant, then why use `.get` with a possible wrong error message instead of keeping the existing code which also relied on this invariant by indexing correctly?

I agree that the invariant is there, but then if we break the invariant, we shouldn't return wrong error messages right?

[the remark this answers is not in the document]
gilcu3: So you propose to introduce a new error variant for the situation that happens if the invariant is broken at some point? 

>why use .get with a possible wrong error message

the only reason is to avoid the use of index slicing :smiling_face_with_tear:

src/crypto/polynomials.rs
DSharifi: It's probably neligible if it's in the order of participants 10~100, depending on how often this is called.

src/crypto/polynomials.rs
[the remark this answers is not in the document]
gilcu3: > I don't have a good solution to the problem. One alternative we could use is using `take` instead of `.get` as `.take` doesn't fail if the amount is more than what's in the iterator.
> 
> Also to find that index, we can use `rposition` isntead
> 
> ```rust
> let len = coefcommitments
>         .iter()
>         .rposition(|x| x.value() != C::Group::identity())
>         .map(|i| i + 1)
>         .unwrap_or(0);
> 
>     if len == 0 {
>         return Err(ProtocolError::EmptyOrZeroCoefficients);
>     }
> 
>     let new_coefficients: Vec<_> = coefcommitments
>         .iter()
>         .take(len)
>         .cloned() 
>         .collect();
> 
>     Ok(Self {
>         coefficients: new_coefficients,
>     })
> ```

That does not look that bad, will make a variant of it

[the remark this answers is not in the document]
DSharifi: > So you propose to introduce a new error variant for the situation that happens if the invariant is broken at some point?

I'm saying don't return a misleading error if an internal invariant is broken.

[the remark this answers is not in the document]
gilcu3: fixed in https://github.com/near/threshold-signatures/pull/264/commits/eb97f8a90ff1d54d587e749c3d925f3fae845678

DSharifi (approved):

PR #282, opened 2026-01-16
fix: remove indexing-slicing exception from robust-ecdsa

### LINKED ISSUE #281: remove indexing-slicing clippy exception from robust-ecdsa
### Background

Currently we have an indexing-slicing clippy exception for the presignature module in robust-ecdsa. We should remove it.

### User Story

_No response_

### Acceptance Criteria

The exception is removed from the robust-ecdsa implementation

_No response_

gilcu3:
Closes #281

### DIFF
--- a/src/ecdsa/robust_ecdsa/presign.rs
+++ b/src/ecdsa/robust_ecdsa/presign.rs
@@ -1,6 +1,3 @@
-// TODO(#122): remove this exception
-#![allow(clippy::indexing_slicing)]
-
 use super::{PresignArguments, PresignOutput};
 use crate::participants::{Participant, ParticipantList, ParticipantMap};
 use crate::{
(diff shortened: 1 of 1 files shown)


src/ecdsa/robust_ecdsa/presign.rs
```diff
@@ -172,21 +169,31 @@ async fn do_presign(
         .into_vec_or_none()
         .ok_or(ProtocolError::InvalidInterpolationArguments)?;
 
+    let (threshold_plus1_identifiers, _) = identifiers
+        .split_at_checked(threshold + 1)
```
netrome: Nit^2: If I'm going to be very nitpicky, `threshold + 1` should be done with a checked addition. But I don't see any realistic scenario where we'll be even close to have a problem with this one so I'm okay being pragmatic here if you feel that would clutter the code too much.

netrome (approved):
Thanks!

src/ecdsa/robust_ecdsa/presign.rs
gilcu3: I think we still have this +1s in several places, so for the time being we can leave it there. We could enable some clippy feature in the future to capture them all at once, will see if I can find one.

EDIT: this is the one https://rust-lang.github.io/rust-clippy/master/index.html#arithmetic_side_effects

Issue #255, opened 2025-12-11
Inconsistent threshold semantics across DKG and presigning protocols


gilcu3:
### Background

This issue corresponds to finding 10 in the Robust ECDSA audit.

The threshold parameter has inconsistent semantic meanings across the DKG and the
robust ECDSA implementation. In the DKG protocol, the threshold represents the
reconstruction bound—that is, the minimum number of participants required to
reconstruct the secret. However, in the robust ECDSA protocol, the threshold parameter
represents the maximum number of malicious parties directly, as explicitly documented in
the warning preceding the do_presign function

### Acceptance Criteria

Short term, standardize the threshold semantics across all protocols in the library and add
documentation to clearly indicate the convention used.

Long term, consider introducing type-safe threshold parameters through distinct types
such as ReconstructionThreshold and MaxMaliciousParties that cannot be
confused or used interchangeably. Add compile-time checks or runtime assertions that
verify that the relationship between DKG and presigning thresholds matches the security
requirements, failing early if mismatched parameters are detected.

### Resources & Additional Notes

We hit this problem already when integrating robust ECDSA in the node, [thread](https://nearone.slack.com/archives/C0912BTG51T/p1765447377296829?thread_ts=1765374628.931179&cid=C0912BTG51T)

PR #310, opened 2026-01-28
Strong typing for DKG

SimonRastikian:
Partially treats #255 
This PR introduces a Strong typing for all the functions related to the generic DKG and testing.
Closing #255  will be done in multiple smaller PRs to allow the reviewer from easily reviewing.
Follow up PRs will deal with strong typing the individual schemes.
See closing comment by netrome in  #306 for more info.

### DIFF
--- a/src/thresholds.rs
+++ b/src/thresholds.rs
@@ -8,13 +8,13 @@ pub struct ReconstructionLowerBound(usize);
 
 // ----- MaxMalicious conversions -----
 impl MaxMalicious {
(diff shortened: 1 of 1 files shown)


netrome:
> @netrome absolutely we can. I still wanted to make all the public function incl. the testing respect this strong typing.

Definitely, but we can skip the `Send + Copy + 'static` bounds if we do this. But no problem if we keep them either. Those bounds should hold for any reasonable type anyway.

netrome (approved):

PR #311, opened 2026-01-28
feat: add strongly typed threshold for all schemes

SimonRastikian:
This is a succeeding PR for  #310 
This deals with the EdDSA and RedDSA schemes.
Follow up for #255

### DIFF
--- a/src/ecdsa/robust_ecdsa/test.rs
+++ b/src/ecdsa/robust_ecdsa/test.rs
@@ -69,7 +69,7 @@ pub fn run_sign_without_rerandomization(
 /// rerandomizing the presignatures
 pub fn run_sign_with_rerandomization(
     participants_presign: &[(Participant, PresignOutput)],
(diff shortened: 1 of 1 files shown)


src/ecdsa/robust_ecdsa/mod.rs
```diff
@@ -16,7 +17,7 @@ pub struct PresignArguments {
     /// This is of type `KeygenOutput`<Secp256K1Sha256> from Frost implementation
     pub keygen_out: KeygenOutput,
     /// The desired threshold for the presignature, which must match the original threshold
-    pub threshold: usize,
+    pub threshold: MaxMalicious,
```
gilcu3: are we keeping the `threshold` name for any reason? Ultimately it should be `max_malicious` no?

gilcu3 (dismissed):
Thank you!

gilcu3 (dismissed):

